PR #5, opened 2018-07-06
Feature/mmt 1507

cgokey:
Couple issues.

I had to increase the timeout as the validation checker was taking longer in some cases.
The ruby method was not raising any errors if the script returned nil or empty, so I modified that and created a custom error, need feedback in how I did this. There are still some records that won't load, but that is due to something wrong in the python script, but atleast it states that now, rather than just a general error (I'll provide a list of records that still won't load and they'll need to modify their python scripts.
The way the script is called, it passes the entire json metadata to python script. The argument is passed in double quotes "", but if there are double anywhere in the json field values, that would cause issues parsing the json. So I stripped out the "" from appearing in any field values, maybe a better way is to escape?
Finally, there was still issues passing the json in double quotes and escaping the double quote " used in the json (which is how it was done here), so I modified it to pass the json in single quotes and escaped the occasional single quote that may appear in the json. This seemed to fix the issues with the other set of records not loading.

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -101,7 +101,7 @@ class CollectionsController < ApplicationController
       rescue => ex
         Rails.logger.error
         ("PyCMR Error: Unknown error ingesting Revision #{params[:revision_id]} with Concept ID #{params[:concept_id]} with error\n#{ex.backtrace}")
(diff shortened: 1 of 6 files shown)


cgokey:
I ran rake test, no new failures, so going to merge.

PR #9, opened 2018-07-09
Feature/mmt 1509

cgokey:
modified code so you can now delete/ import a different granule for in progress/open records.


cgokey:
I removed the old branch and created this one of dev (and not the other ticket).   Based on last review feedback, I've added comments so we remember we need to discuss changing process flow, allowing a user to open a record again after it is closed (this way you can delete/replace the granule).  I also temporarily commented out the test that checks for the ability to be able to delete/replace an granule when the record is closed.

PR #13, opened 2018-07-12
MMT-1482 - downgraded font-awesome-sass from 5.0.13 to 5.0.9 it had a…

cgokey:
downgraded font-awesome-sass from 5.0.13 to 5.0.9 it had a depency that breaks when deployed to SIT environment (sassc), this version uses sass which seems to work.   I think one of these libraries are deprecated so we will need to eventually get off this, but first Redhat systems need to be upgraded to support this, as gcc version is old.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,7 +72,7 @@ gem 'aws-sdk', '~> 2.2.37'
 #store env vars in the .env file
 gem 'dotenv-rails', '~> 2.1.1'
 
-gem 'font-awesome-sass'
+gem 'font-awesome-sass', '5.0.9'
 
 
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -72,7 +72,7 @@ gem 'aws-sdk', '~> 2.2.37'
 #store env vars in the .env file
 gem 'dotenv-rails', '~> 2.1.1'
 
-gem 'font-awesome-sass'
+gem 'font-awesome-sass', '5.0.9'
```
jceaser: add a note that 5.0.13 is not compatible with our servers

PR #15, opened 2018-07-17
Feature/mmt 1509

cgokey:
This fix always keeps the “ingest new granule” button appearing.   arcteam told us this ia a critical feature that needs to be pushed to production.


jceaser:
Paul said off line that he is good with this change.

PR #16, opened 2018-07-18
MMT-1510 Make "Awaiting Release to DAAC" visible to ARC curators

htranho:


jceaser:
This looks pretty strait forward, do we want to do a pull request review on it tomorrow?

cgokey:
I think so… Worth understanding why it is just that one line.

Chris

PR #20, opened 2018-07-27
added cache-control and pragma response headers.

cgokey:
added cache-control and pragma response headers.


jceaser:
we are not going to do this right now, Keven said this can be a very low priority.

PR #21, opened 2018-07-30
MMT 1472 - added active record session store.

cgokey:
Session store now using active record.


config/initializers/session_store.rb
```diff
@@ -1,3 +1,3 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: '_cmr-metadata-review_session'
+Rails.application.config.session_store :active_record_store, :key => '_my_app_session'
```
cgokey: Use the original key

cgokey:
use the whenever gem to trim sessions.

config/initializers/session_store.rb
```diff
@@ -1,3 +1,3 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: '_cmr-metadata-review_session'
+Rails.application.config.session_store :active_record_store, :key => '_cmr-metadata-review_session'
```
charleshuang80: you should keep the syntax as `key:` (they're equivalent but this one is preferred if it is possible)

config/schedule.rb
```diff
@@ -0,0 +1,6 @@
+# Use this file to easily define all of your cron jobs.
+#
+set :output, "/tmp/cron_log.log"
+every 1.day, at: '4:00 am' do
+  rake "db:sessions:trim"
```
charleshuang80: This rake task and the cron job don't output anything if it runs successfully. So for MMT Hoan created a different rake task that invokes this one and also adds something to the logs so we know it was run successfully. We probably want to do the same thing here.

config/schedule.rb
```diff
@@ -0,0 +1,6 @@
+# Use this file to easily define all of your cron jobs.
+#
+set :output, "/tmp/cron_log.log"
```
cgokey: Just noticed, I thought I took out the set output, I probably shouldn't write to /tmp, there is another ticket to in the queue for handling writing logs to splunk, but is still open.   I'll look at what Hoan has done interim though.

PR #26, opened 2018-08-10
CMRARC-342 Searching for specific concept id yields 0 results

cgokey:
The reason it was returning 0 results was due to how the query was formulated.     The CMR search would tag ?* on to the end of the query.   This resulted in a query requiring an extra character added to the end.   Removing the ? fixed the issue.


app/models/cmr.rb
```diff
@@ -417,9 +417,9 @@ def self.collection_search(free_text, provider = ANY_DAAC_KEYWORD, curr_page = "
 
       #setting the two versions of free text search we want to run (with/without first char wildcard)
       options = base_options.clone
-      options["keyword"] = "?*#{free_text}?*"
+      options["keyword"] = "*#{free_text}*"
```
jceaser: this should be trimmed, the extra spaces do affect the results

app/models/cmr.rb
```diff
@@ -417,9 +417,9 @@ def self.collection_search(free_text, provider = ANY_DAAC_KEYWORD, curr_page = "
...
       options_first_char = base_options.clone
-      options_first_char["keyword"] = "#{free_text}?*"
+      options_first_char["keyword"] = "#{free_text}*"
```
jceaser: we should log when this is used, how often is this query used.

test/models/cmr_test.rb
```diff
@@ -79,13 +79,27 @@ class CmrTest < ActiveSupport::TestCase
...
+      stub_request(:get, "https://cmr.earthdata.nasa.gov/search/collections.echo10?page_num=1&page_size=10&provider=NSIDCV0&provider=ORNL_DAAC&provider=LARC_ASDC&provider=LARC&provider=LAADS&provider=GES_DISC&provider=GHRC&provider=SEDAC&provider=ASF&provider=LPDAAC_ECS&provider=LANCEMODIS&provider=NSIDC_ECS&provider=OB_DAAC&provider=CDDIS&provider=LANCEAMSR2&provider=PODAAC&keyword=*doris*").with(:headers => {'Accept' => '*/*', 'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'User-Agent' => 'Ruby'}).to_return(:status => 200, :body => "HTTP/1.1 200 OK
+Access-Control-Allow-Origin: *
```
charleshuang80: I think this was just supposed to be the response body, but you pasted the response status and headers in here too...

app/models/cmr.rb
```diff
@@ -407,6 +407,7 @@ def self.collection_search(free_text, provider = ANY_DAAC_KEYWORD, curr_page = "
     collection_count = 0
 
     if free_text
+      free_text = free_text.strip
```
charleshuang80: I think this is fine as is. But just FYI (in case you didn't already know) for Ruby methods there are often 'bang' (`!`) alternatives that modify the object itself instead of returning a modified copy. see the two versions of `strip` here https://ruby-doc.org/core-2.5.1/String.html#method-i-strip

jceaser: rubocop does not complain

jceaser:
I'll do a merge for Chris since he is out this week.

PR #41, opened 2018-10-02
Feature/cmrarc 394

cgokey:
preprocesses xml from CMR and removes empty tags.

### DIFF
--- a/app/models/cmr.rb
+++ b/app/models/cmr.rb
@@ -155,11 +155,11 @@ class Cmr
     format_collection(raw_collection, data_format)
   end
 
-  def self.remove_empty_tags(node, path)
+  def self.remove_empty_tags(node)
     # post order traversal of DOM to remove leaf nodes that have no content
     # i.e., <Instruments/>
     node.children.each do |child|
-      remove_empty_tags(child, path)
+      remove_empty_tags(child)
       # if is a leaf node and content is empty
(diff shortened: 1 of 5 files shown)


app/models/cmr.rb
```diff
@@ -249,7 +267,13 @@ def self.get_raw_collection(concept_id, type = "echo10")
     data = Cmr.cmr_request(url).parsed_response
 
     begin
-      collection_results = type == "umm_json" ? JSON.parse(data) : Hash.from_xml(data)["results"]
+      if type == "umm_json"
+        collection_results = JSON.parse(data)
+      else
+        doc = Nokogiri.XML(data)
```
cgokey: Refactor the Nokogiri to one call (maybe if)

app/models/record_formats/echo10_record.rb
```diff
@@ -99,6 +99,12 @@ def evaluate_script(raw_data = nil)
         comment_hash = Record.format_script_comments(comment_hash, value_keys)
         comment_hash
       else
+        if (collection?)
+          identifier = "Collection #{short_name} #{version_id}"
+        else
+          identifier = "Granule #{self.get_column("GranuleUR")}"
+        end
+        Rails.logger.error("Python error occurred (#{identifier})")
```
cgokey: look into passing error message as argument to PythonError class

PR #42, opened 2018-10-11
Feature/cmrarc 355

cgokey:
These are changes necessary for earth data login along with some tests.


app/models/user.rb
```diff
@@ -1,8 +1,9 @@
 class User < ActiveRecord::Base
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable and :omniauthable
-  devise :database_authenticatable, :registerable,
-         :recoverable, :rememberable, :trackable, :validatable, :lockable
+  # :database_authenticatable, :registerable,
+  # :recoverable, :rememberable, :trackable, :validatable, :lockable
```
cgokey: Remove these comments

Gemfile
```diff
@@ -77,6 +77,11 @@ gem 'font-awesome-sass', '5.0.9'
 gem 'font-awesome-rails'
 gem 'activerecord-session_store'
 gem 'whenever', require: false
+# gem 'oauth'
+gem 'oauth2'
+# gem 'omniauth'
```
cgokey: Remove commented out ones

Gemfile
```diff
@@ -77,6 +77,11 @@ gem 'font-awesome-sass', '5.0.9'
...
+gem 'omniauth-oauth2'
+gem 'omniauth-urs', :git => "https://git.earthdata.nasa.gov/scm/cmrarc/omniauth-urs.git"
```
cgokey: Create a tag fro gem and reference this to prevent future breakage, see MMT.

Gemfile
```diff
@@ -86,4 +91,5 @@ group :development do
   gem 'spring'
 
   gem 'rdoc'
+  gem 'httplog'
```
cgokey: Drop this

app/views/shared/_navigation.html.erb
```diff
@@ -37,21 +37,20 @@
...
-          <%= link_to "#", class: "menu_link header_button" do %>
-            <i class="fa fa-sliders"></i>
-            <p>Account Options</p>
-          <% end %>
-          <ul class="eui-sub-menu">
-            <li><%= link_to "Edit Account", edit_user_registration_path %></li>
-            <% if current_user.admin %>
-              <li><%= link_to "Create User", new_user_registration_path %></li>
-              <li id="record_refresh"><%= link_to "Refresh Records", record_refresh_path %></li>
+        <% if current_user.admin %>
```
cgokey: Maybe add ?

config/routes.rb
```diff
@@ -2,13 +2,18 @@
...
 
+    get '/login', to: 'login#urs'
+
+    # https://github.com/plataformatec/devise/issues/1390
     devise_scope :user do
       root to: "devise/sessions#new"
-      get '/users/sign_out', to: 'users/sessions#destroy'
+      delete "/users/sign_out" => "devise/sessions#destroy"
+      get '/session/new(.:format)' => 'devise/sessions#new', as: :new_user_session
+      get '/session/sign_out(.:format)' => 'devise/sessions#destroy', as: :destroy_user_session
```
cgokey: Look to see if I need this.

test/controllers/login_controller_test.rb
```diff
@@ -0,0 +1,63 @@
...
+      mock_auth_hash
+      request.env["devise.mapping"] = Devise.mappings[:user] # If using Devise
+      request.env['omniauth.auth'] = OmniAuth.config.mock_auth[:urs]
+    end
+
+    describe "#urs callback" do
+      it "should successfully create a user" do
+        mock_auth_hash
+        before = User.count
+        puts "#{User.count}"
```
cgokey: remove

test/controllers/login_controller_test.rb
```diff
@@ -0,0 +1,63 @@
...
+    before do
+      mock_auth_hash
+      request.env["devise.mapping"] = Devise.mappings[:user] # If using Devise
+      request.env['omniauth.auth'] = OmniAuth.config.mock_auth[:urs]
+    end
+
+    describe "#urs callback" do
+      it "should successfully create a user" do
+        mock_auth_hash
+        before = User.count
```
cgokey: before_count

app/views/shared/_navigation.html.erb
cgokey: So I'm going to leave this the way it is, as the comment above admin function says, think maybe create a separate ticket as technical debt to refactor this.
  # ==== Method
  # This method checks if the user is a "admin", which is a legacy role name. It's only
  # intended for backwards compatibility. For new code, use the 'role' attribute directly.
  def admin
    role.eql?("admin")
  end

app/views/devise/sessions/new.html.erb
```diff
@@ -11,39 +11,18 @@
...
-        <div class="form_buttons">
-          <%= f.submit("Log in", class: 'selectButton') %>
-        </div>
-      </div>
-    <% end %>
-  </div>
-
-  <%= render "devise/shared/links" %>
-  
+  <%- if devise_mapping.omniauthable? %>
```
charleshuang80: are these dashes (`-`) in these erb tags serving a purpose? if not, can we remove them?

cgokey: Looks like they are not needed, interesting discussion on all the variations though:
https://stackoverflow.com/questions/7996695/what-is-the-difference-between-and-in-erb-in-rails

config/initializers/devise.rb
```diff
@@ -236,9 +236,8 @@
   config.sign_out_via = :delete
 
   # ==> OmniAuth
-  # Add a new OmniAuth provider. Check the wiki for more information on setting
-  # up on your models and hooks.
-  # config.omniauth :github, 'APP_ID', 'APP_SECRET', scope: 'user,public_repo'
+  # Add a new OmniAuth provider.
+  config.omniauth :urs, ENV['CLIENT_ID'], ENV['CLIENT_SECRET']
```
charleshuang80: I'm not sure if this is the right place, but I think with both this PR and the gem, we need to address the issue we discussed of connecting to the appropriate URS environment (SIT, UAT, PROD), as this code moves through our environments.

db/schema.rb
```diff
@@ -124,6 +124,8 @@
     t.integer "failed_attempts", default: 0, null: false
     t.string "unlock_token"
     t.datetime "locked_at"
+    t.string "provider"
+    t.string "uid"
```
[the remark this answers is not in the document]
cgokey: This is standard naming convention for omniauth.  Also it includes the provider column which combined makes it unique.   provider could be "urs" or "saml", so you'd distinguish it that way.

config/initializers/devise.rb
cgokey: Another ticket has been created for this.

app/views/shared/_navigation.html.erb
cgokey: I added this to technical debt..   
https://bugs.earthdata.nasa.gov/browse/CMRARC-412

Gemfile
```diff
@@ -77,6 +66,10 @@ gem 'font-awesome-sass', '5.0.9'
 gem 'font-awesome-rails'
 gem 'activerecord-session_store'
 gem 'whenever', require: false
+gem 'oauth2'
+gem 'omniauth-oauth2'
+
+gem 'omniauth-urs', :git => "https://git.earthdata.nasa.gov/scm/cmrarc/omniauth-urs.git", branch: "develop"
```
jceaser: remind me why we are going to pull from develop and not master? Should the default not pull from master and the active developer switch to develop? Or is this just till we merge develop into master at the end of the sprint and from then on it will be master by default?

app/models/user.rb
```diff
@@ -57,4 +54,30 @@ def daac_curator?
...
     admin || curator
   end
+
+  def self.from_omniauth(auth)
+
+    user = User.where(:provider => auth.provider, :uid => auth.uid).first
+    if !user
+      user = User.new
+      user.uid = auth.uid
+      user.role = 'arc_curator'
```
jceaser: is this an "enum" somewhere? what are the full possibilities of values?

Gemfile
cgokey: This is just temporary, I think we want to create a tag, when things are mature and working as expected.

app/models/user.rb
cgokey: they are not enum but I like the idea of a technical ticket moving in this direction though.   The full possibilities are arc_curator, daac_curator, and admin

cgokey: Interesting, the closest it looks ruby has to enum is symbols...  But the value side of symbols can't be stored in a DB, I don't believe, so we'd have to change this implementation to use constants.  https://stackoverflow.com/questions/75759/enums-in-ruby

cgokey: What I'd do is include in the other refactoring ticket for roles, the use of constants, so this doesn't get lost.

app/views/devise/sessions/new.html.erb
```diff
@@ -11,39 +11,18 @@
...
-        <% end -%>
-        </div>
-        <div class="form_buttons">
-          <%= f.submit("Log in", class: 'selectButton') %>
-        </div>
-      </div>
+  <% if devise_mapping.omniauthable? %>
+    <% resource_class.omniauth_providers.each do |provider| %>
+      <%
+        display_name = OmniAuth::Utils.camelize(provider);
```
charleshuang80: what is a 'provider' here, and how many do we expect?
Either way, this type of code should be done in a view helper.

db/migrate/20180829195404_add_omniauth_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddOmniauthToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :provider, :string
+    add_column :users, :uid, :string
```
charleshuang80: I don't see my previous comment - I had suggested that this be `urs_uid` if there's a chance of other user ids being incorporated.

test/controllers/login_controller_test.rb
```diff
@@ -0,0 +1,28 @@
...
+Dir[Rails.root.join("test/**/*.rb")].each {|f| require f}
+
+# https://gist.github.com/kinopyo/1338738
+class LoginControllerTest < ActionController::TestCase
+  include OmniauthMacros
+
+  describe "POST #urs" do
+    before do
+      mock_auth_hash
+      request.env["devise.mapping"] = Devise.mappings[:user] # If using Devise
```
charleshuang80: not a huge deal, but this comment seems to be either unclear or unnecessary to me.

cgokey: Agree, removed it.

app/views/devise/sessions/new.html.erb
cgokey: The provider in this case is "urs" or "saml", it is basically the omniauth strategy name. I'll talk to you about view helper and purpose.

cgokey: I've included a helper method for the business logic that deals with camelizing the string.

app/views/devise/sessions/new.html.erb
```diff
@@ -11,39 +11,18 @@
...
-      </div>
+  <% if devise_mapping.omniauthable? %>
+    <% resource_class.omniauth_providers.each do |provider| %>
+      <%
+        display_name = camelize_str(provider);
+        if provider == :urs
+          display_name = "Earthdata Login"
+        end
+      %>
+      <%= link_to "Login with #{display_name}", omniauth_authorize_path(resource_name, provider) %><br/>
```
charleshuang80: Sorry if I wasn't clear, I thought you might want to move the whole `display_name` logic (i.e. move lines 19 - 24 into the helper), and on line 25 have the `link_to "Login with #{display_name(provider)}"`...
I might also try to move the whole iterating through providers logic into the helper, but I'm not sure if that's feasible.

PR #43, opened 2018-10-11
Removed configuration value to not perform mail deliveries 

cgokey:
Note this is a pull request into master since it is a hot fix.   
Removed configuration value to not perform mail deliveries and just to be sure commented out the code in the mailer class that actually performs the mailing.


cgokey:
hot fix for disabling mailing.

PR #49, opened 2018-10-25
CMRARC-410 now uses cmr base url for access to cmr repository.

cgokey:


app/models/collection.rb
```diff
@@ -93,7 +93,10 @@ def self.create_new_record_by_url(url, user = nil)
   end
 
   def self.parse_collection_url(url)
-    match_data = url.match(/https:\/\/cmr\.earthdata\.nasa\.gov(:443)?\/search\/concepts\/(C\d*-.*)\/(\d*)\.(.*)/)
```
cgokey: Add comment to what this is returning.

cgokey: Added a collection unit test to test this method

PR #53, opened 2018-11-06
CMRARC-407 Functionality added to derive the role from ACLs

cgokey:
Functionality added to retrieve ACLs from CMR and derive the dashboard's role (i.e, arc_curator, daac_curator, or admin).


app/models/user.rb
```diff
@@ -56,15 +56,17 @@ def arc?
   end
 
   def self.from_omniauth(auth)
-
     user = User.where(:provider => auth.provider, :uid => auth.uid).first
     if !user
       user = User.new
       user.uid = auth.uid
-      user.role = 'arc_curator'
```
cgokey: This should be removed

app/models/user.rb
```diff
@@ -56,15 +56,17 @@ def arc?
...
+      user.role = Cmr.getRole(auth.uid, auth.credentials["access_token"]);
       user.provider = auth.provider
```
cgokey: Just make a comment of what this provider is.

app/models/acl_dao.rb
```diff
@@ -0,0 +1,91 @@
...
+    @client_id = client_id
+    @base_url = base_url
+  end
+
+  def get_role(user_id)
+
+    # Sets whether the user is a specific role type
+    # Note: user could be all three.
+    user_is_daac_curator = false
+    user_is_arc_curator = true
```
charleshuang80: should this be `true`? does this mean they automatically get this role?

app/models/acl_dao.rb
```diff
@@ -0,0 +1,91 @@
...
+    user_is_admin = false
+
+    results = search_acls_by_user(user_id, 1)
+    noPages = (results['hits']/2000).ceil
+    items = results['items']
```
charleshuang80: I think you should use the `Array.wrap` here, and remove it from line 25

app/models/acl_dao.rb
```diff
@@ -0,0 +1,91 @@
...
+    if (items.nil?)
+      return nil
+    end
+
+    items = Array.wrap(items)
+    for i in 2..noPages+1
```
charleshuang80: in Ruby, `for... in...` loops are allowed but discouraged. Prefer `each` or `each_with_index`

app/models/acl_dao.rb
```diff
@@ -0,0 +1,91 @@
...
+      results = search_acls_by_user(user_id, i)
+      items += Array.wrap(results['items'])
+    end
+
+    items.each {|item|
```
charleshuang80: single line blocks should use the brackets `{ }`, multi line blocks should use `do ... end`

app/models/acl_dao.rb
```diff
@@ -0,0 +1,91 @@
...
+      if item['name'].end_with?"DASHBOARD_ADMIN"
+        user_is_admin = true;
+      end
+    }
+
+    #
+    # returns the highest level role
+    #
+    role = nil
+    if (user_is_admin)
```
charleshuang80: this statement looks a bit odd. i think you can use `case` (I think it checks them in order but you should verify) or use `elsif` instead of further nesting the `if` statements

app/models/acl_dao.rb
cgokey: Yes, that was my logic at the time, but I've been rethinking this today and going to talk with Alicia, but I think given anyone can get a earth data login account, we shouldn't default to arc curator like the existing application does.   In fact, if they have no ACLs, we should block access.   I'm going to confirm this but I think that is probably the way we should go.

app/models/acl_dao.rb
cgokey: Agree makes sense.

app/models/acl_dao.rb
cgokey: Changed.

app/models/acl_dao.rb
cgokey: I modified this quite a bit, think it reads a little more clear.

app/models/acl_dao.rb
```diff
@@ -0,0 +1,87 @@
...
+      if item['name'].end_with?"DASHBOARD_ADMIN"
+        roles << "admin"
+      end
+    }
+
+    #
+    # returns the highest level role
+    #
+
+    if (roles.include? "admin")
```
charleshuang80: 1. I thought `arc_curator` was a 'higher' role than `daac_curator`, since the former is a system level permission?
2. these `if` statements are short enough that if you want to keep them separate you should make put them on one line (and remove the parens https://github.com/rubocop-hq/ruby-style-guide#no-parens-around-condition), so `return <role> if <role check>` but...
3. you can still link these together with `elsif` I think the one line `if` statements might be fewer lines so that _might_ be preferable

app/models/user.rb
```diff
@@ -56,15 +56,17 @@ def arc?
   end
 
   def self.from_omniauth(auth)
```
charleshuang80: I believe class methods should be declared above instance methods (https://github.com/rubocop-hq/ruby-style-guide#consistent-classes)

app/controllers/login_controller.rb
```diff
@@ -2,13 +2,16 @@ class LoginController < Devise::OmniauthCallbacksController
 
   def urs
     @user = User.from_omniauth(request.env["omniauth.auth"])
-
-    if @user.persisted?
+    if @user.persisted? && !@user.role.nil?
```
charleshuang80: maybe add some comments to explain a bit more what is happening here

app/models/user.rb
charleshuang80: http://www.railstips.org/blog/archives/2009/05/11/class-and-instance-methods-in-ruby/

app/models/acl_dao.rb
cgokey: #1 No, arc curators lower, arc curators release records to daacs and daacs are allowed to close records.  Hopefully they won't have multiple ACL roles but if you are part of a daac, you should be given this privilege.      I changed the rest to reflect the new (to me) way of doing the if statements.

app/models/acl_dao.rb
```diff
@@ -0,0 +1,72 @@
...
+    items.each do |item|
+      roles << "daac_curator" if item['name'].end_with?"DASHBOARD_DAAC_CURATOR"
+      roles << "arc_curator" if item['name'].end_with?"DASHBOARD_ARC_CURATOR"
+      roles << "admin" if item['name'].end_with?"DASHBOARD_ADMIN"
+    end
+
+    #
+    # returns the highest level role
+    #
+    return "admin" if (roles.include? "admin")
```
charleshuang80: the parens around the roles check should be removed https://github.com/rubocop-hq/ruby-style-guide#no-parens-around-condition

app/models/acl_dao.rb
```diff
@@ -0,0 +1,72 @@
...
+  def get_role(user_id)
+
+    # Sets whether the user is a specific role type
+    # Note: user could be all three.
+
+    results = search_acls_by_user(user_id, 1)
+    noPages = (results['hits']/2000).ceil
+    items = Array.wrap(results['items'])
+
+    (2..noPages+1).each { |i|
```
charleshuang80: the brackets `{ }` here for the `each` block still need to be changed to `do` and `end`

PR #56, opened 2018-11-07
Feature/earth data login

cgokey:


cgokey:
Merging earth data login changes to dev.

PR #57, opened 2018-11-09
CMRARC-418 - Testing revealed a few issues.  1) If the user is a

cgokey:
Testing revealed a few issues.  1) If the user is a
daac_curator, an associated daac should be assigned.  Modified the
ACL code to retrieve the daac associated with the daac curator and both are
now assigned to the user.   Note it uses the provider as the daac.

2) Added tests for pulling in daac curators.   Also if it is a daac
added a test to make sure by checking the page you log into, daac curators
don't have a section called "Unreviewed" only arc curators and admins.
Tested for this for their use cases as well.

3) omniauth-urs code was updated in the developer branch and for the gem was being cached, so had to rebuild by clearing the cache on the build machine.

### DIFF
--- a/app/models/acl_dao.rb
+++ b/app/models/acl_dao.rb
@@ -7,7 +7,8 @@ class AclDao
     @base_url = base_url
   end
 
-  def get_role(user_id)
+  # returns [role, daac_name]
+  def get_role_and_daac(user_id)
 
     # Sets whether the user is a specific role type
     # Note: user could be all three.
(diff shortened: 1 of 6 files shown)


app/models/acl_dao.rb
```diff
@@ -22,18 +23,24 @@ def get_role(user_id)
     end
 
     roles = Set.new
+
+    # this logic assumes that a user can only belong to 1 daac, which is the case with dashboard.
+    daac = nil
     items.each do |item|
-      roles << "daac_curator" if item['name'].end_with?"DASHBOARD_DAAC_CURATOR"
+      if item['name'].end_with?"DASHBOARD_DAAC_CURATOR"
```
cgokey: We should log if for some reason a user is provisioned with multiple DAACs.
Check with Alicia on wording for user not provisioned.

test/features/can_access_top_page_test.rb
```diff
@@ -42,6 +42,7 @@ class CanAccessTopPageTest < Capybara::Rails::TestCase
         click_link "Login"
         page.must_have_content("Logout")
         page.must_have_content("Account Options")
+        page.must_have_content("Unreviewed Records:")
```
cgokey: Add in tests for other sections.

PR #62, opened 2018-11-16
Hotfix/cmrarc 419

cgokey:
Since we setup a new provider called ARCTEST, this code includes this in the list of providers to search, so these records can be found.  Note I didn't change the PROVIDER list as I don't want ARCTEST showing up in the list of providers available in the search page or in the metrics.

### DIFF
--- a/app/models/cmr.rb
+++ b/app/models/cmr.rb
@@ -408,35 +408,35 @@ class Cmr
 
     if free_text
       free_text = free_text.strip
-      base_options = {"page_size" => page_size, "page_num" => curr_page}
+      base_options = {'page_size' => page_size, 'page_num' => curr_page}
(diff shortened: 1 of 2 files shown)


app/models/cmr.rb
```diff
@@ -412,6 +412,7 @@ def self.collection_search(free_text, provider = ANY_DAAC_KEYWORD, curr_page = "
       #setting the provider params
       if provider == ANY_DAAC_KEYWORD
         base_options["provider"] = PROVIDERS
+        base_options["provider"] << "ARCTEST"
```
htranho: single quote here?

cgokey: yes, will make the change.

PR #64, opened 2018-11-16
Dev

jceaser:
merging in the dev changes to support earthdata login.


jceaser:
this code was reviewed when it went from feature branch to dev, this is the final review

PR #74, opened 2018-12-03
CMRARC-414 code to check and ensure account is active.

cgokey:
omniauth has a "active_for_authentication?" method that is called on each request.   This code uses that method to make a call to URS for account info, if we get a response, we know that the account is still active.   If it returns anything but code 200, we present a message saying the account is no longer active.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,6 +12,8 @@ class User < ActiveRecord::Base
 
   def self.from_omniauth(auth)
     user = User.where(:provider => auth.provider, :uid => auth.uid).first
+    user.access_token = auth.credentials["access_token"]
+
     if !user
       user = User.new
       user.uid = auth.uid
@@ -34,6 +36,25 @@ class User < ActiveRecord::Base
     else
       super
     end
+  end
+
+  def active_for_authentication?
+    super && check_if_account_active
+  end
+
+  def check_if_account_active
+    conn = Faraday.new(:url => "#{ENV['urs_site']}") do |faraday|
+      faraday.request :url_encoded # form-encode POST params
+      faraday.headers['Authorization'] = "Bearer #{self.access_token}"
+      faraday.response :logger # log requests to $stdout
+      faraday.adapter Faraday.default_adapter # make requests with Net::HTTP
+    end
+    response = conn.get "/api/users/#{uid}?calling_application=#{ENV['urs_client_id']}"
+    return response.status == 200
+  end
+
+  def inactive_message
+    "Sorry, this account has been deactivated."
   end
 
   # ====Params
--- a/db/migrate/20181203171821_add_access_token_to_users.rb
+++ b/db/migrate/20181203171821_add_access_token_to_users.rb
@@ -1,0 +1,5 @@
+class AddAccessTokenToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :access_token, :string
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180829195404) do
+ActiveRecord::Schema.define(version: 20181203171821) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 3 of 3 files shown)


app/models/user.rb
```diff
@@ -36,6 +40,46 @@ def self.new_with_session(params, session)
     end
   end
 
+  def active_for_authentication?
```
charleshuang80: I'm not seeing where this is being set to run before every page/controller action - is it baked into devise?

app/models/user.rb
```diff
@@ -36,6 +40,46 @@ def self.new_with_session(params, session)
...
+    super && check_if_account_active
+  end
+
+  def check_if_account_active
+    return true if Rails.env.test?
```
charleshuang80: after thinking about this a bit, I think that this should be removed here because we should have a test to make sure this test works. to make the tests work we can mock/stub `get_user_info` or move this line into that method.

app/models/user.rb
```diff
@@ -36,6 +40,46 @@ def self.new_with_session(params, session)
...
+    end
+    response.status == 200
+  end
+
+  def inactive_message
+    "Sorry, this account has been deactivated."
+  end
+
+  def refresh_access_token
+    conn = Faraday.new(:url => "#{ENV['urs_site']}") do |faraday|
```
charleshuang80: I am not sure if it needs to be resolved in this PR, but I am wondering if http/api calls should be made in the model (I think more likely it is a no), or if there's a better place to put these http/api calls.

app/models/user.rb
```diff
@@ -80,4 +124,17 @@ def daac_curator?
   def arc?
     admin || curator
   end
+
+  private
+
+  def get_user_info
```
charleshuang80: same question about a http/api call in the model. I think methods in models should have tests. I also think that api calls (in general but not necessarily for all cases) should be be mocked, so if this method is kept in the model we can either stub it or move your `return true if Rails.env.test?` line here.

app/models/user.rb
cgokey: As mentioned in a side conversation, it is called by devise.

app/models/user.rb
cgokey: Maybe this would be good to pair over the IP sprint this part.     If I mock get_user_info, the check_if_account_active is called by lots of tests, where is the best place to mock this (without being redundant in code?)
(See the latest code, changed a bit)

app/models/user.rb
cgokey: Yeah I'd like to as discussed move this logic into the before_action but was having some trouble, so thinking will revisit during IP sprint.

app/models/user.rb
cgokey: Same response as above...

app/models/cmr.rb
```diff
@@ -632,5 +632,35 @@ def self.get_role_and_daac(uid, access_token)
...
+      faraday.headers['Authorization'] = 'Basic ' + ["#{ENV['urs_client_id']}:#{ENV['urs_client_secret']}"].pack('m0')
+      faraday.response :logger # log requests to $stdout
+      faraday.adapter Faraday.default_adapter # make requests with Net::HTTP
+    end
+    response = conn.post "/oauth/token",
+                         grant_type: "refresh_token",
+                         refresh_token: current_user.refresh_token
+
+    json = JSON.parse(response.body)
+    [json["access_token"], json["refresh_token"]]
```
cgokey: Mimic to see what happens if an error occurs

cgokey: Log all external calls to Rails.logger

app/models/user.rb
```diff
@@ -36,6 +40,30 @@ def self.new_with_session(params, session)
     end
   end
 
+  def active_for_authentication?
```
cgokey: Log errors if not 200

test/controllers/login_controller_test.rb
```diff
@@ -1,34 +1,46 @@
...
 # https://gist.github.com/kinopyo/1338738
 class LoginControllerTest < ActionController::TestCase
   include OmniauthMacros
+
   describe "POST #urs" do
     before do
       mock_auth_hash
+      ENV['urs_site'] = 'https://sit.urs.earthdata.nasa.gov'
+      ENV['urs_client_id'] = 'clientid'
+      ENV['urs_client_secret'] = 'clientsecret'
```
cgokey: I should refactor to my common module method

PR #82, opened 2018-12-18
CMRARC-347 - adding a status check interface

jceaser:
this will check if the database is working, similar to how MMT does things

### DIFF
--- a/app/middleware/middleware_healthcheck.rb
+++ b/app/middleware/middleware_healthcheck.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: false
+
+# create a status checkup interface
+class MiddlewareHealthcheck
+  def initialize(app)
+    @app = app
+  end
+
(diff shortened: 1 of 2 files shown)


app/middleware/middleware_healthcheck.rb
```diff
@@ -0,0 +1,56 @@
...
+
+# create a status checkup interface
+class MiddlewareHealthcheck
+  def initialize(app)
+    @app = app
+  end
+
+  # this is the Middleware handler
+  def call(env)
+    if env['PATH_INFO'.freeze] == '/status'.freeze
```
jceaser: suggestion to move freeze to the outside of the brackets.

jceaser: it does not make since to me anymore to freeze the result of the lookup, I'm just removing it all together.

PR #85, opened 2018-12-18
Feature/cmrarc 420

cgokey:
Changes to present user name and role to user.


app/assets/stylesheets/pages/_landing-page.scss
```diff
@@ -172,11 +172,40 @@ body {
   height: 100%;
 }
 
-.landing_page {
-}
+#user-info {
```
cgokey: Move this to _userinfo_box.scss ?

app/views/shared/_navigation.html.erb
```diff
@@ -7,6 +7,13 @@
   </div>
 
   <div class="header_menu">
+    <section id="user-info">
+      <div class="prof-link-container">
```
charleshuang80: I don't think shaving off a few letters saves you much here but can add a bit of confusion, I think the class should just be `profile-link-container`

app/models/user.rb
```diff
@@ -26,6 +26,10 @@ def self.from_omniauth(auth)
     role, daac = Cmr.get_role_and_daac(auth.uid, auth.credentials['access_token'])
     user.role = role
     user.daac = daac if daac
+    auth.info['first_name'] = '' if auth.info['first_name'].nil?
+    auth.info['last_name'] = '' if auth.info['last_name'].nil?
```
charleshuang80: `auth.info` is what's coming back from URS via omniauth, right? I don't think you should be changing what is coming back from that, it might be better to just declare new variables for first and last names.

charleshuang80: Also, first and last names are required fields when you register with URS. I'm not sure you need these checks in that case, especially if you are using interpolation within double quotes, and then doing a `strip`.

app/models/user.rb
```diff
@@ -26,6 +26,10 @@ def self.from_omniauth(auth)
...
+    user.name = auth.info['first_name']+' '+auth.info['last_name']
```
charleshuang80: I think rubocop should have yelled at you about this - you should use interpolation in double quotes here instead of using `+' '+` to concatenate strings.

app/assets/stylesheets/application.scss
```diff
@@ -26,6 +26,7 @@
 @import "layout/footer";
 @import "components/masthead";
 @import "components/general";
+@import "components/userinfo_box";
```
charleshuang80: since 'user info' is actually two words, I think this file should actually be `_user_info_box`

app/assets/stylesheets/pages/_landing-page.scss
lhlyles: Updated to components/_user_info_box.scss

app/views/shared/_navigation.html.erb
lhlyles: This is the same class name used in MMT

app/assets/stylesheets/application.scss
lhlyles: Changed name to _user_info_box as recommended

PR #90, opened 2019-01-18
CMRARC-444 initial prototype implementation of tracking granule revisions.

cgokey:
Some goals:
  Be able to add a new granule review
   Be able to delete any granule review
   Be able to detect if a granule has been deleted in CMR
      example: C1279404075-GES_DISC
   Be able to detect if a new granule revision comes available in CMR
      example: C1000000362-LPDAAC_ECS
   Be able to pull in that new granule revision


app/controllers/granules_controller.rb
```diff
@@ -6,6 +6,45 @@ class GranulesController < ApplicationController
   def show
   end
 
+  # Pulls a new random granule and associates with the specified collection id.
```
cgokey: Clarify that this isn't changing CMR but rather working with records, reviews ,etc.

app/controllers/granules_controller.rb
```diff
@@ -6,6 +6,45 @@ class GranulesController < ApplicationController
...
+  def create
+    collection = Collection.find(params[:id])
+    authorize! :create_granule, collection
+    collection.add_granule_with_concept_id(current_user, collection.concept_id)
+    flash[:notice] = collection.granules.count > 0 ?  "A new random granule has been added for this collection" : "No Granules Available!"
+    redirect_to collection_path(id: 1, record_id: collection.records.first.id)
+  end
+
+  # Deletes the specified granule record (and its associated review) using the
+  # specified granule id.
```
cgokey: should say "granule record id"

app/helpers/collections_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+    highest_revision_id = 0
+    granule.records.each do |granule_record|
+      revision_id = granule_record.revision_id.to_i
+      highest_revision_id = revision_id if revision_id > highest_revision_id
+    end
+    if highest_revision_id < granule.latest_revision_in_cmr.to_i
+      return link_to("Import New Revision #{granule.latest_revision_in_cmr}",
+                     pull_latest_granule_path(granule.id),
+                     method: :post, class: 'import_new_revision', id: 'ingest_pull_latest',
+                     data: { confirm: 'Are you sure?' })
```
cgokey: Should we use EUI's modal confirm dialog?

app/models/granule.rb
```diff
@@ -74,6 +74,39 @@ def self.add_granule_by_concept_id(granule_concept_id, current_user = User.find_
...
+          order_count:  i,
+          daac:         daac_from_concept_id(granule.concept_id)
+        )
+        granule_record.save!
+      end
+
+
+      granule_ingest = Ingest.create(record: granule_record, user: current_user, date_ingested: DateTime.now)
+      granule_ingest.save!
+      granule.save!
```
cgokey: Consider not using !

cgokey: Capture and return back if it doesn't not save and why.

app/models/granule.rb
```diff
@@ -74,6 +74,39 @@ def self.add_granule_by_concept_id(granule_concept_id, current_user = User.find_
...
+      true
+    end
+  rescue Cmr::CmrError
+    false
```
cgokey: Only rescues from CmrErrors

PR #91, opened 2019-01-29
Feature/cmrarc 379

cgokey:
cleaned up code, added more error checkings, more comments, few more tests, moved stub
 content (like echo records, dif records) to files instead of inline, etc.


app/models/ability.rb
```diff
@@ -23,6 +23,8 @@ def initialize(user)
         can :close, Record
 
         can :replace_granule, Granule
+        can :delete_granule, Granule
+        can :create_granule, Collection
```
cgokey: So I probably also need to think about what options are available to the view, at minimum provide these options to admin, but check what options should be available for other user types (curator, daac)

app/models/collection.rb
```diff
@@ -124,46 +126,76 @@ def add_granule(current_user)
...
-            granules_components = Granule.assemble_granule_components(self.concept_id, granules_count, self, current_user)
+
+    #only selecting granules for certain formats per business rules
+    if Collection::INCLUDE_GRANULE_FORMATS.include? native_format
+      #creating all the Granule related objects
+      granules_components = Granule.assemble_granule_components(concept_id, granules_count, self, current_user)
+    end
+
+    #saving all the related collection and granule data in a combined transaction
+    granules_components.flatten.each {|savable_object|
```
cgokey: So I looked into this why save! is being called and it seems like at one time they tried to do a transaction to handle all the saves being done under one transaction, but then for some reason pulled this out.   You can check git history to see that.   I'd like to check and see if there is any comments why they took it out.

app/assets/stylesheets/application.scss
```diff
@@ -26,6 +26,7 @@
 @import "layout/footer";
 @import "components/masthead";
 @import "components/general";
+@import "components/confirm_box";
```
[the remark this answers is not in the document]
cgokey: So I added this but getting some strange behavior where the click event works the first time but than once I click no, the click event is no longer registered with the button.   Can demo it on Monday.

PR #94, opened 2019-03-20
CMRARC-468 - Updated a Gem to the lattest supported in its series

jceaser:
Updating the Devise Gem.


jceaser:
I don't think we need to review this in person.

jceaser:
Bamboo is green for the branch

PR #100, opened 2019-07-21
CMRARC-480 fixed issue with some collections and granules not ingesting properly.

cgokey:
fixed issue with some collections and granules not ingesting properly.


app/models/cmr.rb
```diff
@@ -279,7 +279,9 @@ def self.get_raw_collection(concept_id, type = "echo10")
     data = Cmr.cmr_request(url).parsed_response
 
     begin
-      collection_results = convert_xml_to_hash(type, data)["results"]
+      collection_results = convert_xml_to_hash(type, data)
+       # dif and echo return results subelement but umm does not
+      collection_results = collection_results['results'] unless collection_results['results'].nil?
```
cgokey: is there a format field I can check

app/models/cmr.rb
```diff
@@ -412,6 +414,8 @@ def self.random_granules_from_collection(collection_concept_id, granule_count =
       granule_data_list = []
 
       #getting a random list of granules addresses within available amount
+      # CMR limits the page depth to 1,000,000, see ticket CMRARC-480
+      total_granules = [1000000, total_granules].min
```
cgokey: does it include the max number?

ryanmiller-1: ... excludes max
https://ruby-doc.org/core-2.1.3/Range.html

PR #101, opened 2019-07-21
CMRARC-462 handles linking granule revisions to collections.

cgokey:


app/controllers/records_controller.rb
```diff
@@ -19,6 +19,37 @@ def refresh
...
+      flash[:notice] = "This revision #{@record.revision_id} associated granule is marked as 'Granule Review Deleted!"
+      @record.associated_granule_value = associated_granule_value
+    else
+      granule_record = Record.find_by id: associated_granule_value
+      if !granule_record.nil?
+        granule = Granule.find_by id: granule_record.recordable_id
+        granule_concept_id = granule.concept_id
+        granule_revision_id = granule_record.revision_id
+        flash[:notice] = "Granule #{granule_concept_id}/#{granule_revision_id} has been successfully associated to this collection revision #{@record.revision_id}. "
+        @record.associated_granule_value = granule_record.id
```
charleshuang80: does the 'association' happen before or after this assignment, and before or after the `save!` below?

cgokey: Not sure I understand, it happens before, as the save! is after in code, but I'm sure you mean something else.

test/helpers/user_helpers.rb
```diff
@@ -10,5 +10,19 @@ def mock_login(id: nil, uid: nil, role: nil)
       ApplicationController.any_instance.stubs(:user_signed_in).returns(true)
       ApplicationController.any_instance.stubs(:current_user).returns(user)
     end
+
+    def wait_for_jQuery(secs = Capybara.default_max_wait_time)
+      ActiveSupport::Notifications.instrument 'mmt.performance', activity: 'Helpers::AjaxHelpers#wait_for_jQuery' do
```
cgokey: rename this

test/test_helper.rb
```diff
@@ -27,19 +27,11 @@
     # This makes javascript console logs available, but doesn't cause them to appear in real time
     # to display javascript logs in the rspec output, add `puts page.driver.browser.manage.logs.get(:browser)`
     # in the desired test location
-    loggingPrefs: { browser: 'ALL', client: 'ALL', driver: 'ALL', server: 'ALL' }
-  )
+    loggingPrefs: { browser: 'ALL', client: 'ALL', driver: 'ALL', server: 'ALL' },
+    chromeOptions: { args: %w[headless disable-gpu no-sandbox --window-size=1500,1500], w3c: false}
```
cgokey: w3c: false is needed for retrieving javascript console messages.

app/controllers/records_controller.rb
cgokey: The latest code should be much cleaner here.

PR #102, opened 2019-07-23
CMRARC-491 mitigated omniauth security vulnerability until official fix comes available

cgokey:


Gemfile.lock
```diff
@@ -1,11 +1,10 @@
-GIT
-  remote: https://git.earthdata.nasa.gov/scm/cmrarc/omniauth-urs.git
-  revision: 552d4faba52ff0a0727cd170cdaaead138d2ab74
-  tag: release/v1.0.0
+PATH
+  remote: /Users/cgokey/src/omniauth-urs
```
jceaser: this can't be correct

README.md
```diff
@@ -1,7 +1,13 @@
 # CMR Metadata Review
-
 The CMR Metadata Review tool is used to curate NASA EOSDIS collection and granule level metadata in CMR for correctness, completeness and consistency.
 
+## Security Warning
```
jceaser: I think this should come out since we don't actually see the warning banner unless your apart of the github team.

Gemfile.lock
cgokey: yeah don't approve this yet then was trying to get ahead of it, approve the other one first and I'll create a release then it should update properly.

app/views/shared/_navigation.html.erb
```diff
@@ -53,7 +53,7 @@
         <% end %>
 
         <li>
-          <%= link_to destroy_user_session_path, class: "menu_link header_button" do %>
+          <%= link_to destroy_user_session_path, method: :post, class: "menu_link header_button" do %>
```
jceaser: this should be get, not post

PR #104, opened 2019-07-24
CMRARC-493 - added policy to block external scripts unless whitelisted

jceaser:


config/application.rb
```diff
@@ -25,6 +25,17 @@ class Application < Rails::Application
     
     config.middleware.insert_after "Rails::Rack::Logger", "MiddlewareHealthcheck"
 
+    policies = [
+      #'default-src ' + ['\'self\''].join(' '),
+      'script-src-elem ' + ['\'self\'',
+        "\'unsafe-inline\'",
+        #"\'nonce-view012345\'",
```
cgokey: do you want to keep the 2 commented lines in here?

charleshuang80: second the question

jceaser: yes, but I should document what they are

PR #109, opened 2019-08-08
feature/CMRARC-460 - pie charts now include assoc granules

cgokey:


app/controllers/reports_controller.rb
```diff
@@ -54,6 +54,18 @@ def selection
 
   def review
     @records = Record.where(id: params[:record_id])
+    @granule_associations = {}
+    @records.each do |record|
+      value = record.associated_granule_value
+      if value.nil? || value == 'Undefined'
```
charleshuang80: I think there was a question if value could possibly be an empty string. If it is possible, `.nil?` won't catch that, but `.blank?` will.

cgokey: Well it isn't possible for that value to be empty, but regardless I didn't know that, I thought nil? checked for empty as well, so will change it, good to know thanks.

app/views/reports/_review_record.erb
```diff
@@ -0,0 +1,178 @@
...
+              <td><%= record.revision_id %></td>
+              <td><%= record.short_name %></td>
+              <td><%= record.daac %></td>
+              <td><%= link_to "UMM", record.umm_json_link %></td>
+              <td><%= record.format %></td>
+              <td><%= record.formatted_closed_date %></td>
+            </tr>
+          </table>
+
+          <br>
```
charleshuang80: I think it would be better to add some css to the table and add extra space to the bottom

app/models/acl_dao.rb
```diff
@@ -93,7 +93,12 @@ def send_request_to_cmr(action, endpoint, data = nil)
     msg += "#{@base_url}/#{endpoint}, response.status=#{response.status}, "
     msg += "body=#{response.body}"
     Rails.logger.info(msg)
-    json = JSON.parse(response.body)
+    json = {}
+    begin
+      json = JSON.parse(response.body)
+    rescue JSON::ParserError
+      Rails.logger.info "Error parsing JSON response from CMR retrieving ACLs.  response=#{response.body}"
+    end
```
ryanmiller-1: Are the callers of this prepared to handle and understand the empty json?  Is it possible they're proceeding with a failure silently rather than dying because they received an irreconcilable parse error?

app/views/reports/review.html.erb
```diff
@@ -17,182 +17,21 @@
...
-              <% record.reviews.each do |review| %>
-                <tr>
-                  <td><%= review.reviewer_email %></td>
-                  <td><%= review.formatted_date %></td>
-                  <td><%= review.report_comment %></td>
-                </tr>
-              <% end %>
-            </table>
-          </div>
+      <br>
```
charleshuang80: same with these `<br>` tags

test/controllers/login_controller_test.rb
```diff
@@ -45,7 +44,7 @@ class LoginControllerTest < ActionController::TestCase
         before_count = User.count
         post :urs, provider: :urs
         after_count = User.count
-        assert(after_count.must_equal(before_count + 1));
+        # assert(after_count.must_equal(before_count + 1));
```
charleshuang80: does this mean nothing is actually being tested here?

app/models/acl_dao.rb
cgokey: Well this is a one-off, but I see your point.  I have another error message I can use instead that is a bit more generic.

cgokey: Oh wait, they actually don't see this error message, that is what goes to Rails logger, this should forces it to return {} which the calling class parses and presents an error since it got nothing legible from CMR.

ryanmiller-1: That's more of what I was trying to ask about.  It seemed like the old way was failing loudly when it broke and this way had the potential to be failing quietly, but you say it's covered.

app/controllers/reports_controller.rb
```diff
@@ -54,6 +54,18 @@ def selection
 
   def review
     @records = Record.where(id: params[:record_id])
+    @granule_associations = {}
+    @records.each do |record|
+      value = record.associated_granule_value
+      if value.nil? || value.empty? || value == 'Undefined'
```
charleshuang80: You can leave this as is, but if the `.empty?` check is not needed I think it can be removed.

test/controllers/login_controller_test.rb
charleshuang80: speaking with Chris this is commented out to be addressed later.

I think it might be nice to add a comment/TODO to that effect.

cgokey: it is commited and pushed now thanks

PR #110, opened 2019-08-08
Feature/cmrarc 463 - changes to incorporate csv reporting of assoc granules

cgokey:
added csv reporting of assoc granules


app/models/multi_record_csv.rb
```diff
@@ -1,64 +1,72 @@
...
         FORMATS.each do |metadata_format|
           records_for_format = collections.metadata_format(metadata_format)
           next if records_for_format.empty?
 
-          fields = determine_fields(records_for_format)
+          granules = []
+          # Get a list of all granules
+          records_for_format.each do |record|
+            associated_granule_value = record.associated_granule_value
+            next unless !associated_granule_value.nil? &&
```
charleshuang80: This is kind of confusing because you are using `unless` and a lot of `!`s. Is it possible to use `if` and make it a bit more straightforward?

app/views/reports/review.html.erb
```diff
@@ -5,7 +5,7 @@
     <div class="report_drop_down">
     </div>
     <div class="report_buttons">
-      <%= link_to reports_review_path(format: :csv, record_id: @records.map(&:id)), class: "report_button" do %>
+      <%=link_to reports_review_path(format: :csv, record_id: @records.map(&:id)), data: { turbolinks: false }, class: "report_button" do %>
```
charleshuang80: might want to add back the space after `<%=`

app/models/multi_record_csv.rb
cgokey: I originally had if with if but ruby cop changed it :)

jceaser: maybe because next in assumed to be the natural state and rubocop wants to focus on the exception? I vote for rebelling against our robot overlords and change it to if

cgokey: alright I should probably also change it from nil? to empty? anyhow

app/models/multi_record_csv.rb
```diff
@@ -33,9 +33,11 @@ def to_csv
           # Get a list of all granules
           records_for_format.each do |record|
             associated_granule_value = record.associated_granule_value
-            next unless !associated_granule_value.nil? &&
-                        associated_granule_value != 'Undefined' &&
-                        associated_granule_value != 'No Granule Review'
+            if associated_granule_value.empty?        ||
```
jceaser: this section looks much better to my eyes

charleshuang80: agree

charleshuang80: However, I'm unclear if the `associated_granule_value` can be nil. If you want to catch both nil and empty, you should use `.blank?`, but if it can only be one but not the other, then you can use the single appropriate check

PR #112, opened 2019-08-12
CMRARC-490: Edit Collection in MMT link only visible to Daac Curators

charleshuang80:
* add tests
* some css modifications for tests and best practices

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ All development work is done on local copies of the code base.  Follow the steps
 
 - Some people may find they need to create an initial database to connect to with `/usr/local/opt/postgresql@9.6/bin/createdb -h localhost`
 
-4. From the PSQL console, you will need to add a `SUPERUSER` matching the username values shown in `config/database.yml` within the project repo.  The development & test DB's both are in local Postgres.  There are many ways to create a postgres db and superuser.  These links can help –  [here](https://launchschool.com/blog/how-to-install-postgresql-on-a-mac) and [here](https://www.postgresql.org/docs/9.1/static/sql-createrole.html).
+4. From the PSQL console, you will need to add a `SUPERUSER` matching the username values shown in `config/database.yml` within the project repo. The development & test DB's both are in local Postgres. There are many ways to create a postgres db and superuser.  These links can help –  [here](https://launchschool.com/blog/how-to-install-postgresql-on-a-mac) and [here](https://www.postgresql.org/docs/9.1/static/sql-createrole.html).
 
 - With user information from `config/database.yml`, the commands will look like: `CREATE ROLE <user> WITH SUPERUSER PASSWORD '<password>';`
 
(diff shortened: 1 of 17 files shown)


app/models/user.rb
```diff
@@ -82,34 +82,25 @@ def records_not_reviewed
...
-  # None
-  # ====Returns
-  # Whether or not the User is an admin.
-  # ==== Method
+  # TODO: are these comments still valid?
   # This method checks if the user is a "admin", which is a legacy role name. It's only
   # intended for backwards compatibility. For new code, use the 'role' attribute directly.
   def admin
-    role.eql?("admin")
+    role == 'admin'
```
cgokey: it seems to me that admin, curator, and daac_curator can all be removed in favor of. using the role property in the model.   maybe before they had some other way for determining if the user was an admin, then finally created a role property?

test/features/can_show_collections_test.rb
```diff
@@ -5,30 +5,85 @@ class CanShowCollectionsTest < Capybara::Rails::TestCase
...
+    end
+  end
+
+  describe 'when the user is a Daac Curator' do
+    before do
+      OmniAuth.config.test_mode = true
+      # TODO: looking at the uids in user.yml, this user's provider doesn't seem
+      # to match the provider of the collection that shows up in screenshots,
+      # but this is the uid that works.
+      mock_login(role: 'daac_curator', uid: 5)
```
cgokey: So uid:5 is podaac, and doesn't what shows up is record 
```
podaac_in_daac_review:
  id: 14
  recordable_id: 2
  recordable_type: "Collection"
  revision_id: "7"
  state: "in_daac_review"
  format: "dif10"

```
which points to collection.yml #2 which is a podaac record?

cgokey: wait.. record_data.yml I think determines the daac the collection belongs to.   and that is:
```
in_daac_review_short_name:
  id: 20
  record_id: 12
  column_name: "ShortName"
  value: "A2_DySno_NRTb"
  daac: "PODAAC"
  last_updated: null
  color: "green"
  script_comment: "ok"
  opinion: false
  flag: []
  recommendation: "ok"


in_daac_review_long_name:
  id: 21
  record_id: 12
  column_name: "LongName"
  value: "Test_Long_Name"
  daac: "PODAAC"
  last_updated: null
  color: "green"
  script_comment: "ok"
  opinion: false
  flag: []
  recommendation: "ok"


in_daac_review_version_id:
  id: 22
  record_id: 12
  column_name: "VersionId"
  value: "0"
  daac: "PODAAC"
  last_updated: null
  color: "white"
  script_comment: "ok"
  opinion: true
  flag: []
  recommendation: "ok"

```

app/models/user.rb
charleshuang80: If I were starting from scratch I would want to have the role property, but I would also probably implement these helper methods, especially the `arc?` method. The one thing I would change might be for these two methods to change to `admin?` and `arc_curator?`
So do you think we should remove them or maybe make the change and keep them?

cgokey: think just adding the ? makes sense.   seems like a nice convenience method to just user.admin? and there maybe other logic in the future that might want to check for these kinds of things.

test/features/can_show_collections_test.rb
charleshuang80: The concept id and short name that shows up (it's not letting post the screenshot) is actually this one from the collections.yml file:
```
first:
  id: 1
  concept_id: "C1000000020-LANCEAMSR2"
  short_name: "A2_DySno_NRT"
```

app/models/user.rb
charleshuang80: I will make the change for the helper methods.

When searching to make the changes, it looks like the Users table used to have an admin field and a curator field, before they added the role field.

README.md
```diff
@@ -39,7 +39,7 @@ All development work is done on local copies of the code base.  Follow the steps
 
 - Some people may find they need to create an initial database to connect to with `/usr/local/opt/postgresql@9.6/bin/createdb -h localhost`
 
-4. From the PSQL console, you will need to add a `SUPERUSER` matching the username values shown in `config/database.yml` within the project repo.  The development & test DB's both are in local Postgres.  There are many ways to create a postgres db and superuser.  These links can help –  [here](https://launchschool.com/blog/how-to-install-postgresql-on-a-mac) and [here](https://www.postgresql.org/docs/9.1/static/sql-createrole.html).
+4. From the PSQL console, you will need to add a `SUPERUSER` matching the username values shown in `config/database.yml` within the project repo. The development & test DB's both are in local Postgres. There are many ways to create a postgres db and superuser.  These links can help –  [here](https://launchschool.com/blog/how-to-install-postgresql-on-a-mac) and [here](https://www.postgresql.org/docs/9.1/static/sql-createrole.html).
```
jceaser: I can't tell you how happy I am to see the old double space typewriter legacy go away

PR #115, opened 2019-08-14
Feature/cmrarc 454b -functionality to revert a record

cgokey:


test/controllers/records_controller_test.rb
```diff
@@ -121,4 +121,34 @@ class RecordsControllerTest < ActionController::TestCase
...
+      record = Record.find(12)
+      assert_equal Record::STATE_IN_DAAC_REVIEW.to_s, record.state
+      post :revert, id: record.id
+      record = Record.find(12)
+      assert_equal Record::STATE_READY_FOR_DAAC_REVIEW.to_s, record.state
+    end
+
+  end
+
+
```
ryanmiller-1: Is it worth including a test to verify that you cannot revert any more aspects of the system?

cgokey: Good idea and added a test.

PR #119, opened 2019-09-05
CMRARC-459 adding new specific granule

cgokey:


app/controllers/granules_controller.rb
```diff
@@ -111,4 +111,29 @@ def replace
       redirect_to collection_path(id: 1, record_id: collection.records.first.id)
     end
   end
+
+  def ingest_specific
```
charleshuang80: if the controller action and the model method have the same name, could that create confusion down the road?

app/models/granule.rb
```diff
@@ -107,6 +107,42 @@ def self.add_new_revision_to_granule(granule, current_user = User.find_by(role:
...
 
+  def self.ingest_specific(collection_concept_id, granule_concept_id, current_user = User.find_by(role: "admin"))
+    granule_info = Cmr.get_granule_with_collection_data(granule_concept_id)
+    collection = Collection.find_by(concept_id: granule_info["collection_concept_id"])
+
+    if granule_info["collection_concept_id"] != collection_concept_id
+      raise Cmr::CmrError.new, 'Granule record does not belong to this collection.'
+    end
+    if collection.nil?
+      raise Cmr::CmrError.new, 'Collection record associated with granule could not be found in system.'
```
charleshuang80: I don't know of a better alternative here at the moment, but instead of returning with something that indicates an error, you are raising an exception. I'm not sure if it needs to be changed, but it is propagating the need to use `rescue` in more places that we've discussed.

cgokey: Yeah in other languages that is kind of common, I don't think you want your return to be an error object or something.   Maybe this could be refactored to raise specific exception types (that is more common in other languages), but that might be a little overkill for this use case.

app/controllers/granules_controller.rb
cgokey: I don't think so, but I'm up for suggestions to rename it?

app/controllers/granules_controller.rb
```diff
@@ -111,4 +111,29 @@ def replace
...
+      collection = Collection.find(params[:id])
+      authorize! :create_granule, collection
+      collection_concept_id = collection.concept_id
+      granule_concept_id = params["granule_concept_id"]
+
+      record = collection.granules.find_by concept_id: granule_concept_id
+      unless record.nil?
+        flash[:notice] = "Sorry, granule review #{granule_concept_id} already exists!"
+        redirect_to collection_path(id: 1, record_id: collection.records.first.id)
+        return
```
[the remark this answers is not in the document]
cgokey: I put the "and return" although rubocop complains about using "and" which is kind of strange.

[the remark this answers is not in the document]
charleshuang80: yeah, it's come up before. I think it's more of a rails-y thing that seems a common way to do it

PR #124, opened 2019-09-17
CMRARC-503 - Updating the Devise GEM

jceaser:
This is a basic bundle update, local tests pass


jceaser:
Bamboo is green, merging.

PR #126, opened 2019-09-23
CMRARC-484 - changes to fix auth token.

cgokey:


README.md
```diff
@@ -73,3 +73,15 @@ All development work is done on local copies of the code base.  Follow the steps
 - Register for an [Earthdata Login account](https://sit.urs.earthdata.nasa.gov/) for the SIT environment.
 
 - Request ACL permissions to access the tool in the SIT environment by emailing [Earthdata Support](mailto:support@earthdata.nasa.gov). In order to Ingest collections into the tool, you may need Admin or Arc Curator permissions as well.
+
+## Other Known Issues
+
+During the RAILS 5.2 upgrade, there was an issue with the CSRF authenticity tokens.   Namely, this specific 
+workflow:  if a user clicks See Review Details, then clicks Curation Home, then clicks Revert Record,
+they will get a Invalid Authenticity Token.   Workaround is to tell form_with it should not auto include the
+token, rather we should explicitly include it ourselves. i.e.,
```
charleshuang80: is it worth mentioning here that we only need to include it for non get requests?

test/features/can_revert_record_test.rb
```diff
@@ -0,0 +1,43 @@
...
+      mock_login(role: 'admin')
+
+      stub_request(:get, "#{Cmr.get_cmr_base_url}/search/granules.echo10?concept_id=G309210-GHRC")
+        .with(
+          headers: { 'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Accept' => '*/*', 'User-Agent' => 'Ruby' }
+        )
+        .to_return(status: 200, body: '<?xml version="1.0" encoding="UTF-8"?><results><hits>0</hits><took>32</took></results>', headers: {})
+    end
+
+    it 'can revert a record, mimics workflow in issue CMRARC-484' do
```
charleshuang80: If you are going to reference the issue, I think it is worth commenting that this test would not actually catch the error that needed to be fixed.

app/views/collections/show.html.erb
```diff
@@ -100,7 +100,14 @@
 
             <div id="<%= modal_id %>" class="eui-modal-content confirm-modal">
               <a href="#" class="modal-close float-r"><i class="fa fa-times"></i><span class="is-invisible">Close</span></a>
-              <%= form_tag(ingest_specific_granule_path(:id => @granule_objects.first.collection.id), method: "post") do %>
+              <!-- Todo: Verify AuthenticityToken issue is fixed in RAILS 6.0 -->
```
jceaser: should this not be an ERB comment not an HTML comment:

<%# This is a comment! %>

cgokey: Well there are 71 comments in the ERB files using <!-- so that the seems the convention for this project at least, not sure what is the convention though in general.

jceaser: do the others mention security flaws?

PR #129, opened 2019-10-25
CMRARC-455 - logic to handle preserving recommendations

cgokey:


app/models/record.rb
```diff
@@ -777,4 +777,67 @@ def related_granule_record
...
+  def get_column_record_data_map(record)
+    record_data_map = {}
+    record.record_datas.each do |data|
+      record_data_map[data.column_name] = data
+    end
+    record_data_map
+  end
+
+  # Takes the recommendations from the prior record and copies them into the existing record.
+  # Note: the recommendations include color,feedback, last_updated, opinion, recommendation field
```
cgokey: also includes discussions.

test/features/copy_recommendations_test.rb
```diff
@@ -0,0 +1,73 @@
...
+        end
+        accept_alert do
+          click_on "Copy Prior Recommendations"
+        end
+        page.must_have_content('Successfully copied recommendations')
+      end
+
+      # this use case will copy recommendations from a prior revision but if we try it again, it will verify
+      # the button is disabled so we can't perform the action again.
+      it 'copies recommendations.' do
```
cgokey: make this more descriptive

app/controllers/collections_controller.rb
```diff
@@ -11,7 +11,7 @@ def show
       collection = get_collection_from_record(@record)
 
       @concept_id = collection.concept_id
-      @collection_records = collection.get_records.order(:revision_id).reverse_order
```
cgokey: This logic will not properly order based on revision id since revision_id is a string, it will do a string comparison.

app/models/cmr.rb
```diff
@@ -33,6 +33,14 @@ def self.cmr_request(url)
     end
   end
 
+  def self.get_raw_concept(concept_id, revision_id = nil)
```
cgokey: This code isn't necessary for this ticket but useful library I'm using for some side code that I use for small rake tasks, that I will eventually check in.

app/models/cmr.rb
```diff
@@ -85,6 +93,10 @@ def self.update_collections(current_user)
     end
 
     update_lock = RecordsUpdateLock.find_by id: 1
+    if update_lock.nil?
+      update_lock = RecordsUpdateLock.new
+      update_lock.id = 1
+    end
```
cgokey: This fixes a bug with records refresh

app/models/record.rb
```diff
@@ -204,8 +204,8 @@ def values
   end
 
   def get_previous_revision
-    # a set of fellow records sorted by revision id in ascending order
-    collection_records = self.recordable.records.order(:revision_id).to_a
+    # a set of records sorted by revision id in ascending order
+    collection_records = self.recordable.get_records(false)
```
cgokey: Now uses the fixed version of getting records sorted.

app/views/collections/show.html.erb
```diff
@@ -26,11 +26,11 @@
     <h3 class="collection_revision_header">Collection Revisions</h3>
     <% @collection_records.each_with_index do |record, index| %>
 
-      <div class="collection_revision">
+      <div id="collection_revision_<%= record.revision_id %>" class="collection_revision">
```
cgokey: Mostly creating the "id" for tests so I can access different parts of the page.

app/views/collections/show.html.erb
```diff
@@ -26,11 +26,11 @@
...
         <table cellpadding="5" border="1">
 
           <tr>
-            <th>
+            <th class="collection_granule_revision_row">
```
cgokey: Now uses css that fixes the size of the header, makes it a little more aligned and consistent.

app/views/collections/show.html.erb
```diff
@@ -119,79 +119,79 @@
...
       <% @granule_objects.each_with_index do |granule, index| %>
-        <% granule.records.each do |granule_record| %>
-            <table cellpadding="5" border="1">
-
-              <tr>
-                <th>
-                  <h4>
-                    <div>Concept ID: <%= granule_record.concept_id %>
-                      <span class="indicator_for_granule_deleted_in_cmr"><%= indicator_for_granule_deleted_in_cmr(granule) %></span>
+        <% granule.get_records.each do |granule_record| %>
```
cgokey: This also now grabs a sorted list of records

app/views/records/show.html.erb
```diff
@@ -4,10 +4,22 @@
...
   <div class="outer_flexbox">
+
+
     <!-- container of metadata sections -->
     <div class="section_container outer_flex_item_left">
       <!-- header for sections -->
+
       <div class="section_header">
+        <div class="review_buttons_row">
+          <% if should_show_copy_recommendations_button?(@record) %>
```
cgokey: logic for dealing with adding the copy recommendations button to the page.

db/schema.rb
```diff
@@ -10,123 +10,124 @@
...
+    t.string "revision_id", null: false
     t.datetime "closed_date"
-    t.string   "format",                   default: ""
-    t.string   "state"
-    t.string   "associated_granule_value"
-    t.index ["recordable_type", "recordable_id"], name: "index_records_on_recordable_type_and_recordable_id", using: :btree
+    t.string "format", default: ""
+    t.string "state"
+    t.string "associated_granule_value"
+    t.string "copy_recommendations_note"
```
cgokey: Here is where the field is added.

test/controllers/collections_controller_test.rb
```diff
@@ -29,7 +29,7 @@ class CollectionsControllerTest < ActionController::TestCase
 
       get :show, params: { id: 1, record_id: 1 }
       collection_records = assigns(:collection_records)
-      assert_equal(4, collection_records.length)
+      assert_equal(6, collection_records.length)
```
cgokey: Had to update to do a change in # of fixtures.

test/controllers/login_controller_test.rb
```diff
@@ -29,7 +29,6 @@ class LoginControllerTest < ActionController::TestCase
 
     describe "#urs callback" do
       it "should successfully create a user" do
-        skip "Creating a new user. Todo: figure out transaction rollbacks."
```
cgokey: We can no remove this skipped test, as rails 5 has better support for cleaning up transactions after each test, seems to work fine now.

test/features/associating_granules_test.rb
```diff
@@ -19,23 +19,29 @@ class AssociatingGranulesTest < Capybara::Rails::TestCase
     describe 'associate granules to collections' do
       it 'can assign granule to collection' do
         visit '/home'
-        wait_for_jQuery(5)
 
-        all('#record_id_')[0].click # Selects the first collection record in the list
-        find('#open > div > div.navigate-buttons > input').click # click the See Review Details button
-        first("#associated_granule_value").find("option[value='Undefined']").click # other tests are not cleaning up db, so reset it back manually
-        first("#associated_granule_value").find("option[value='5']").click # Click in select box for what granule to associate
+        within '#open' do
```
cgokey: All changes in this test are to use the within block which makes things easier, no real functionality changes to the tests.   Also with the additional fixtures, scoping using within helps to prevent future breakage.

test/features/copy_recommendations_test.rb
```diff
@@ -0,0 +1,73 @@
+require 'test_helper'
```
cgokey: This is where the tests to handle copying recommendations is.

test/helpers/user_helpers.rb
```diff
@@ -11,14 +11,6 @@ def mock_login(id: nil, uid: nil, role: nil)
       ApplicationController.any_instance.stubs(:current_user).returns(user)
     end
 
-    def wait_for_jQuery(secs = Capybara.default_max_wait_time)
```
cgokey: No longer needs int rails 5

test/helpers/user_helpers.rb
```diff
@@ -11,14 +11,6 @@ def mock_login(id: nil, uid: nil, role: nil)
...
-      ActiveSupport::Notifications.instrument 'cmrdash.performance', activity: 'Helpers::AjaxHelpers#wait_for_jQuery' do
-        Timeout.timeout(secs) do
-          loop until finished_all_jQuery_requests?
-        end
-      end
-    end
-
     def finished_all_jQuery_requests?
```
cgokey: I think I need to remove this too actually, don't think it is used anywhere anymore.

test/models/record_test.rb
```diff
@@ -169,7 +169,7 @@ class RecordTest < ActiveSupport::TestCase
   describe "daac scope" do
     it "will return records for the given DAAC" do
       records = Record.daac("PODAAC")
-      assert_equal 5, records.length
+      assert_equal 7, records.length
```
cgokey: also due to adding fixtures

test/test_helper.rb
```diff
@@ -29,12 +29,16 @@
...
     # w3c: false is needed for retrieving javascript console messages.
     loggingPrefs: { browser: 'ALL', client: 'ALL', driver: 'ALL', server: 'ALL' },
-    chromeOptions: { args: %w[headless disable-gpu no-sandbox --window-size=1500,1500], w3c: false}
+    chromeOptions: { args: %w[headless disable-gpu no-sandbox --window-size=1500,2000], w3c: false}
 
   )
   Capybara::Selenium::Driver.new(app, browser: :chrome, http_client: client, desired_capabilities: capabilities)
 end
 
+Capybara::Screenshot.register_driver(:headless_chrome) do |driver, path|
```
cgokey: Needed in rails 5 capybara to grab screenshots wouldn't work without it.

app/helpers/record_helper.rb
```diff
@@ -61,4 +61,17 @@ def script_class(field)
   def script_text_class(field)
     field[:script] && current_user.arc? ? "script_text" : "no_script_text"
   end
+
+  def record_is_in_progress(record)
+    array = [Record::STATE_OPEN.to_s, Record::STATE_IN_ARC_REVIEW.to_s, Record::STATE_READY_FOR_DAAC_REVIEW.to_s]
```
ryanmiller-1: This can be done without the assignment. [stuff].include?(search stuff)

app/models/cmr.rb
ryanmiller-1: Is this the only place the lock gets used?  Does it need to be updated elsewhere?

cgokey: That the only place it "writes" to it.   It is used in one other location for the purpose of "reading" in order to display the time of the last update on the page.

app/views/collections/show.html.erb
```diff
@@ -26,11 +26,11 @@
     <h3 class="collection_revision_header">Collection Revisions</h3>
     <% @collection_records.each_with_index do |record, index| %>
 
-      <div class="collection_revision">
+      <div id="collection_revision_<%= record.revision_id %>" class="collection_revision">
         <table cellpadding="5" border="1">
```
charleshuang80: Just from what I can see in this file, there are three instances of `<table cellpadding="5" border="1">`. I know you didn't add them but we discussed moving it into css. Maybe you can do it little by little, so add the class and use it for the tables when you happen to be making changes to those views?

app/helpers/record_helper.rb
```diff
@@ -61,4 +61,18 @@ def script_class(field)
   def script_text_class(field)
     field[:script] && current_user.arc? ? "script_text" : "no_script_text"
   end
+
+  def record_is_in_progress(record)
```
charleshuang80: if this is returning a boolean, it might be better to have the method name have a question mark - maybe like `def record_in_progress?(record)`

app/assets/stylesheets/pages/_records.scss
```diff
@@ -350,3 +350,14 @@
...
+
+.copy_prior_revision_note {
+  font-size: smaller;
+  font-style: italic;
+  height: 30px;
+}
+
+.eui-btn--disabled {
+  opacity: 0.65;
+  cursor: not-allowed !important;
```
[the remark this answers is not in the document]
cgokey: The no cursor is a nice visual that is all, the actual disable works.   Let me see if I can figure out how to get rid of !important

app/helpers/record_helper.rb
jceaser: it even sounds like it needs a question mark

cgokey: Always forget about this little '?' nuance, thanks.

charleshuang80:
approved

PR #131, opened 2019-11-06
CMRARC-292 - made url calls have a timeout, fixed so newlines are preserved

cgokey:


lib/CheckerCollection.py
```diff
@@ -101,7 +101,7 @@ def checkDesc(self, val):
 
         for url in urls:
             try:
-                connection = urllib2.urlopen(url)
+                connection = urllib2.urlopen(url, timeout=5)
```
jceaser: maybe the value 5 should be a global configuration.

app/assets/stylesheets/pages/_charts.scss
```diff
@@ -429,3 +429,11 @@
   margin-bottom: 20px;
 }
 
+.report_comment {
+  white-space: pre-wrap;
+}
+
+.recommendation_comment {
```
charleshuang80: since the two classes are using the same styling, I think you can just combine them like:
```
.report_comment, .recommendation_comment {
  white-space: pre-wrap;
}
```

lib/CheckerGranule.py
```diff
@@ -3,6 +3,7 @@
 import socket
 from datetime import *
 
+LINK_CHECK_TIMEOUT = 5
```
charleshuang80: this one has spaces around the equals, the one in the other file does not. I like the spaces better, but I don't know what's better for python and I think your preference should be favored here.

PR #132, opened 2019-11-13
Feature/cmrarc 522- fixed spacing issues with 2 column table

cgokey:


cgokey:
I needed up turning both tables into a flex table, the header and the actual review table.  For the header I had to use flex-grow to take it take the available space.   The other table was pretty standard constraints.   Lot of the changes in the erb are because I had to wrap the other tables in a new container (which is a flex container), so no real functionality changes besides that.

app/assets/stylesheets/pages/_show-collection.scss
```diff
@@ -32,8 +33,15 @@
...
   .collection_revision_header {
-    height: 95px;
-    line-height: 90px;
+    height: 60px;
+    line-height: 60px;
+    vertical-align: middle;
+    font-size: 1em;
+  }
+
+  .granules_revision_header {
```
charleshuang80: these two classes are using the same stylings. I made a comment in the other PR about combining them to reduce duplication.

PR #133, opened 2019-11-25
CMRARC-369 changes to make feedback records show up.

cgokey:


app/views/reviews/_requires_feedback.html.erb
```diff
@@ -31,18 +31,20 @@
...
         </p>
-      </div>  
+      </div>
     </td>
 
     <script type="text/javascript">
-      addFeedbackClickHandler("feedback_cell_<%= title %>", "feedback_check_<%= title %>", "feedback_text_<%= title %>", "feedback_icon_<%= title %>");
-      
+      <% if current_user.daac_curator?  %>
+        addFeedbackClickHandler("feedback_cell_<%= title %>", "feedback_check_<%= title %>", "feedback_text_<%= title %>", "feedback_icon_<%= title %>");
```
charleshuang80: I was confused by this method. It looks like javascript but you're adding it in an erb file. Looking more closely, it actually is js, but just in script tags. I think this should be moved to a javascript file, and not called by erb.

test/features/curator_feedback_test.rb
```diff
@@ -0,0 +1,127 @@
...
+
+    stub_request(:get, "#{Cmr.get_cmr_base_url}/search/granules.echo10?concept_id=G309210-GHRC")
+      .with(
+        headers: { 'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Accept' => '*/*', 'User-Agent' => 'Ruby' }
+      )
+      .to_return(status: 200, body: '<?xml version="1.0" encoding="UTF-8"?><results><hits>0</hits><took>32</took></results>', headers: {})
+
+  end
+
+  # This is the base case there are no feedback records, so this:
```
charleshuang80: these comments seem very helpful!

app/views/reviews/_requires_feedback.html.erb
cgokey: I don't think we can move it to a JS file, as this is rendered multiple times and conditionally called on each render.   We only register the click handler <% if current_user.daac_curator?  %> for that specific cell (notice the arguments being passed in).

ryanmiller-1: I believe that we will need to remove most/all of these in the security headers change.

jceaser: well, I think the solution would be different in that case. If javascript truly knows nothing about the current user's permissions, then you could at least load a js file inside the if statement and have that be the javascript that only runs for your case.

cgokey: What I can do this this... I'll include a boolean in the rendered html, whether it is a daac curator or not.   Then have the JS check the element if true, it will register the click handler, otherwise not.

charleshuang80: doing it this way just feels 'wrong' somehow. I can understand not wanting to resolve it with this ticket. Maybe add a comment and write a ticket to fix it in the future?
2 options come to mind for the data, and you can always add a class or other selector for when you want the click handler - 1) you can pass things to JS through erb files 2) this might make more sense for this case, add whatever data that needs to be passed to as a data attribute to the html.

ryanmiller-1: Can't you make it a click event driven by the result of feedback_cell_class?  That's already doing the permission check.  I feel like JQuery shenanigans should be able to get you the rest of what you need in the JS.

app/views/reviews/_requires_feedback.html.erb
```diff
@@ -1,48 +1,20 @@
...
-
-  </script>
-
   <th class="headcol headcol__user-row">Feedback</th>
   
   <% section_titles.each do |title| %>
     <td class="no_padding column_<%=title%>">
-      <div id="feedback_cell_<%= title %>" class="feedback_cell review-toggle">
+      <div id="feedback_cell_<%= title %>" class="<%= feedback_cell_class(current_user) %>"
+           onclick='feedbackClickHandler("feedback_cell_<%= title %>", "feedback_check_<%= title %>", "feedback_text_<%= title %>", "feedback_icon_<%= title %>", <%= current_user.daac_curator? %>)'>
```
ryanmiller-1: My recollection is hazy, but I think we will be revisiting this line when we need to update for the security header (I think onclicks were regarded as just as bad as script tags).  I would definitely prefer to see it resolved, but I'm still thinking about whether I think it needs to be.

charleshuang80: I think this is ok for now, but it's still adding javascript in the div tag. I would still advocate for writing a ticket to revisit, especially if it is used in a few other places -- I'd be happy to take a stab at it.

cgokey: Thanks, I removed the JS from the erb file but I ran into an issue with registering the onclick via jquery (seemed to be registering it twice, so for every click event, the method would fire twice, tried to track it down but to no avail), so I'll revisit this issue later.  The input tag still has the onclick in there but at least there is no JS.    Please take a look at the latest implemenation.

PR #135, opened 2019-12-04
Feature/cmrarc 369- fixed version field in home table

cgokey:
So it ended up being that the join worked as expected and all the results returned was as expected.   The issue was that the original query to pull in all collection records did this:
`@records = @records.includes(:record_datas).where(record_data: { column_name: ['VersionId', 'Entry_ID/Version', 'Version']}).references(:record_data)`

So this query retrieves all records and includes only record datas where the column name is : `VersionId, Entry_ID/Version, or Version. `

Then in the ERB file it pulled out the version in the table like so:
`<td class="center_text"><%= record.record_datas.first.value%></td>
`
So this rendering makes the assumption that the only record datas included as part of the record is version record data.   When I added feedback data to the query, now we've got record data with feedback data as well as version data.  That is where the error occurred.
To fix this, I simply made a version property in the model and now retrieving the version like so:

> def version
>     data = record_datas.where(column_name: ['VersionId', 'Entry_ID/Version', 'Version']).first
>     data.nil? ? 'n/a' : data.value
>   end

Then I can use that property rather than the assumptions stated above and was free to change the initial query to return all collection records rather than only those with those version fields.  That query now looks like so:
`    # only include collection records
    @records = @records.where(recordable_type: 'Collection').distinct`

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -53,10 +53,7 @@ class ApplicationController < ActionController::Base
     @second_opinion_counts = RecordData.where(record: @records, opinion: true).group(:record_id).count
 
     # Version ID, Version, or Feedback records are the only field we need to render on the home page, so just load that
-    # record_set1 = @records.left_outer_joins(:record_datas).where("record_data.column_name IN (?, ?, ?) or record_data.feedback = ?", 'VersionId', 'Entry_ID/Version', 'Version', true).distinct;
-    record_set1 = @records.left_outer_joins(:record_datas).where("record_data.column_name IN (?, ?, ?), 'VersionId', 'Entry_ID/Version', 'Version'").distinct
-    @records = record_set1
-                 #.or(record_set2)
+    @records = @records.joins(:record_datas).where("record_data.column_name IN (?, ?, ?) OR feedback = ?", 'VersionId', 'Entry_ID/Version', 'Version', true).distinct
   end
 
   private
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -53,9 +53,9 @@ class ApplicationController < ActionController::Base
     @second_opinion_counts = RecordData.where(record: @records, opinion: true).group(:record_id).count
 
     # Version ID, Version, or Feedback records are the only field we need to render on the home page, so just load that
-    record_set1 = @records.left_outer_joins(:record_datas).where(record_data: { column_name: ['VersionId', 'Entry_ID/Version', 'Version']}).distinct
-    record_set2 = @records.left_outer_joins(:record_datas).where(record_data: { feedback: true }).distinct
-    @records = record_set1.or(record_set2)
+    record_set1 = @records.left_outer_joins(:record_datas).where("record_data.column_name IN (?, ?, ?) or record_data.feedback = ?", 'VersionId', 'Entry_ID/Version', 'Version', true).distinct;
+    @records = record_set1
+                 #.or(record_set2)
   end
 
   private
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -55,7 +55,7 @@ class ApplicationController < ActionController::Base
     @records = @records.includes({ingest: :user}, :reviews)
 
     # Version ID, Version, or Feedback records are the only field we need to render on the home page, so just load that
-    @records = @records.includes(:record_datas).where(record_data: { column_name: ['VersionId', 'Entry_ID/Version', 'Version']}).or(@records.includes(:record_datas).where(record_data: {feedback:true})).references(:record_data)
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -49,13 +49,11 @@ def filtered_records
       @records = @records.campaign(params[:campaign])
     end
 
-    # Count Second Opinions here for every record
+    # # Count Second Opinions here for every record
```
cgokey: Not sure how a # got added, will remove.

app/controllers/application_controller.rb
```diff
@@ -49,13 +49,11 @@ def filtered_records
...
     @second_opinion_counts = RecordData.where(record: @records, opinion: true).group(:record_id).count
 
-    @records = @records.includes({ingest: :user}, :reviews)
-
     # Version ID, Version, or Feedback records are the only field we need to render on the home page, so just load that
-    @records = @records.includes(:record_datas).where(record_data: { column_name: ['VersionId', 'Entry_ID/Version', 'Version']}).or(@records.includes(:record_datas).where(record_data: {feedback:true})).references(:record_data)
+    @records = @records.joins(:record_datas).where("record_data.column_name IN (?, ?, ?) OR feedback = ?", 'VersionId', 'Entry_ID/Version', 'Version', true).distinct
```
cgokey: See major comment in pull request, this query gets at the essence of what the original programmer was after and includes feedback records, but I really think the query should just select all collection records instead.  The record class has a field called "recordable_type" which has only 2 values: collection or granule.

app/models/record.rb
```diff
@@ -172,6 +172,9 @@ def get_column(column_name)
     end
   end
 
+  def version
```
cgokey: I added a version property to the record model class to make it easier to pull out the version, the existing implementation in the erb file had a major flaw once I added additional records with feedback=true.

app/views/shared/_records_table.html.erb
```diff
@@ -41,7 +41,7 @@
               <td class="center_text"><%= record.revision_id %></td>
               <td><%= record.recordable.short_name %></td>
               <!-- Only Version ID should be loaded -->
-              <td class="center_text"><%= record.record_datas.first.value%></td>
+              <td class="center_text"><%= record.version%></td>
```
cgokey: Here I'm using the new version property, rather than assuming the first record data MUST be a version and ONLY a version.

test/features/can_revert_record_test.rb
```diff
@@ -22,15 +22,16 @@ class CanRevertRecordTest < Capybara::Rails::TestCase
       visit '/home'
 
       within '#in_daac_review' do
-        check 'record_id_'
+        page.save_and_open_page
+        all('#record_id_')[0].click
```
cgokey: rest of just updates are test classes to account for the fact that some of the test data doesn't have a version data and since I fixed the code to pull all collection records, not just collection records with a version, we are seeing a couple extra rows.  So the test data is just grabbing the first which happens to be the correct one.

app/controllers/application_controller.rb
```diff
@@ -52,10 +52,8 @@ def filtered_records
     # Count Second Opinions here for every record
     @second_opinion_counts = RecordData.where(record: @records, opinion: true).group(:record_id).count
 
-    @records = @records.includes({ingest: :user}, :reviews)
-
-    # Version ID, Version, or Feedback records are the only field we need to render on the home page, so just load that
-    @records = @records.includes(:record_datas).where(record_data: { column_name: ['VersionId', 'Entry_ID/Version', 'Version']}).or(@records.includes(:record_datas).where(record_data: {feedback:true})).references(:record_data)
+    # only include collection records
+    @records = @records.where(recordable_type: 'Collection').distinct
```
charleshuang80: so do you not need to care about records with `feedback: true` anymore? or is it that since this is grabbing all collections those get picked up also?

cgokey: yeah, I found it odd that is how they classified something to be a "collection" if it had one of those fields, so I asked Jeanne, she is not sure either.   The side effect of that approach was that feedback records didn't show up, hence the ticket.  This is a much cleaner way of doing it, this way all collection records show up regardless of if they have a version or not.   There is some history to this, as I think they used to include BOTH granules and collections in this list, then they decided to filter it this way.   Now it is just granules, maybe by preloading version details made it faster, I'm not sure.   The sql query this generates is pretty efficient and it does lazy loading, so I can't imagine it would be much slower.   Only way to tell I guess would be to grab a snapshot of the current db and try it out... ?

cgokey: Also cmr enforces collection have a version, so we "shouldn't" see any unexpected results.

charleshuang80: I don't know for sure, but I can't imagine they will have so many collections that are under review at a time that the efficiency of the different sql queries will make a big difference. I was just checking that the right data was getting retrieved.

PR #143, opened 2020-01-31
Feature/CMRARC-404 Removing Paul's e-mail address from Dashboard e-mails

ryanmiller-1:


app/mailers/curator_mailer.rb
```diff
@@ -2,12 +2,12 @@ class CuratorMailer < ApplicationMailer
 
   def released_records(users, records)
 #    @records = records
-#    mail(from: ENV['mailer_sender'], to: users.map(&:email), subject: 'Metadata Curation Tool Released Records')
```
jceaser: I checked, this value is not in igit or in the bamboo variables list, so I don't think there is anything that needs to be removed outside of this code. The only reference I could find was in application.yml.

db/schema.rb
```diff
@@ -15,13 +15,13 @@
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
-  create_table "collections", force: :cascade do |t|
+  create_table "collections", id: :serial, force: :cascade do |t|
```
jceaser: many of us had a talk in person about this change, we feel it is safe to make as we don't think we will need more then 2^64 indexes.

README.md
```diff
@@ -33,7 +33,7 @@ All development work is done on local copies of the code base.  Follow the steps
 
 - You will then be prompted for your password
 
-- I you run into the issue of `password authentication failed for user` – try the command `psql -U postgres -h localhost` and enter in the password for your computer
+- If you run into the issue of `password authentication failed for user` – try the command `psql -U postgres -h localhost` and enter in the password for your computer.  If that command does not work, you may find it easier to alter the pg_hba.conf file.  The location of this file varies based on where the postgres is trying to run from.  It should be located in the data directory.  Using `ps aux | grep postgres` should list the active processes, one of which should contain something vaguely like: ' /usr/local/opt/postgresql@9.6/bin/postgres -D /usr/local/var/postgresql@9.6'.  The location after the -D is the data directory for the instance that is currently running.  After you alter that file, you will need to restart postgres for the changes to take effect.
```
jceaser: thank you for adding comments to the read me!

db/schema.rb
ryanmiller-1: And while this is a change to the file, this should not represent a change to the db because serial was the default behavior before rails 5.X.  I think this is being explicitly stated as serial because the new default is BIGSERIAL in postgres.

PR #144, opened 2020-02-04
Feature/CMRARC-547 Adding Column to display dates in e-mails

ryanmiller-1:
Adding column for released_to_daac_date to records


app/models/record.rb
```diff
@@ -434,7 +434,16 @@ def script_score
...
-
+  # ===Params
+  # None
+  # ===Returns
+  # None
+  # ===Method
+  # Updates released_to_daac_date to the current time.
+  # Intended to be called in an after block to the aasm transition
+  def update_released_to_daac_date
+    self.released_to_daac_date = Time.zone.now
```
cgokey: This isn't a bad way to handle it.  You should probably handle the case if they revert from "release" back to in_daac_ review and I guess clear the date out.

PR #145, opened 2020-02-11
CMRARC-411 Adding ability for user to specify how often they can receive e-mails.

ryanmiller-1:


app/controllers/user_controller.rb
```diff
@@ -0,0 +1,14 @@
+class UserController < ApplicationController
+  def email_preferences
+    redirect_to root_path unless can?(:update_email_preferences, current_user)
+    @selected = current_user.email_preference
```
cgokey: I was a little confused later in use of the word "selected", wonder if maybe @current_preference, selected makes me think a choice just happened in the gui.

ryanmiller-1: done

app/controllers/user_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+  def email_preferences
+    redirect_to root_path unless can?(:update_email_preferences, current_user)
+    @current_preference = current_user.email_preference
+  end
+
+  def update_email_preferences
+    if can?(:update_email_preferences, current_user) && current_user.save_email_preference(params[:email_preference])
+      redirect_to(email_preferences_path, notice: 'E-mail frequency preference updated.')
+    else
+      redirect_to(email_preferences_path, alert: 'E-mail frequency preference could not be updated.')
```
charleshuang80: should we have some logging if this fails?

PR #152, opened 2020-03-06
CMRARC-546 As a dashboard operator, I would like to run and view metadata curation reports as non-EOSDIS CMR provider

cgokey:
After user login, the application goes into 1 of 2 modes: :mdq_mode or :arc_mode
The mode is determined by the user's role and/or associated daac
The mode is then used to determine how the data is filtered.   If it is :mdq_mode, it is filtered using the MDQ_PROVIDER list.   If it is :arc_mode, it is filtered by using the ARC_PROVIDER list.
The filtering is entirely controlled by 2 methods:
  Records.all_records(application_mode)
  Metadata.all_metadata(application_mode)
Much of the rest of the changes were changing the code to use these filtered methods rather than "all" records.
A feature toggle has been implemented to turn on filtering (mdq_enabled_feature_toggle=true) as we are not yet sure how performance will be impacted by this filtering.


app/models/user.rb
```diff
@@ -8,7 +8,7 @@ class User < ApplicationRecord
 
   validates_presence_of :daac, if: Proc.new { |u| u.role.eql?("daac_curator") }
 
-  ROLES = %w[admin arc_curator daac_curator].freeze
+  ROLES = %w[admin arc_curator mdq_curator daac_curator].freeze
```
jceaser: earthdata_curator maybe, but I don't like mdq

db/schema.rb
```diff
@@ -15,13 +15,13 @@
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
-  create_table "collections", id: :serial, force: :cascade do |t|
```
jceaser: why is the ID being removed from all these tables?

app/models/ability.rb
```diff
@@ -66,6 +66,32 @@ def initialize(user)
         can :view_report_comments, Review
       end
 
+      if user.mdq_curator?
```
ryanmiller-1: I can't recall.  If the goal is to have the MDQ and ARC curators be entirely separated in their capabilities, I think that these also need to check that the item being acted on is from one of the appropriate DAACs.

app/models/record.rb
```diff
@@ -667,6 +674,10 @@ def has_short_name?
     self.short_name != ""
   end
 
+  def daac
```
ryanmiller-1: If your code contains my changes, this should be removed.

app/models/user.rb
ryanmiller-1: The ACL in CMR is called MDQ curator, so changing it to something else would only complicate the matter.

db/schema.rb
ryanmiller-1: This is a difference between Chris and my local environments, I think.  I would prefer not to remove it because I think we should specify it now that the default has changed.

db/schema.rb
```diff
@@ -64,25 +64,26 @@
...
+  create_table "records", force: :cascade do |t|
     t.integer "recordable_id", null: false
+    t.string "recordable_type", null: false
     t.string "revision_id", null: false
     t.datetime "closed_date"
     t.string "format", default: ""
     t.string "state"
     t.string "associated_granule_value"
     t.string "copy_recommendations_note"
+    t.datetime "state_last_modified"
```
ryanmiller-1: searching on the page, I don't see this used or the migration that added this.

db/schema.rb
```diff
@@ -15,13 +15,13 @@
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
-  create_table "collections", id: :serial, force: :cascade do |t|
+  create_table "collections", force: :cascade do |t|
     t.string "concept_id", null: false
```
ryanmiller-1: We discussed migrating daac up to the collection/granule.  Why did you decide not to pursue that?

lib/tasks/fill_test_data.rake
```diff
@@ -10,7 +10,7 @@ task :fill_test_data => :environment do
   record_count = 500
   record_data_count = 10000
 
-  search_iterator, count = Cmr.collection_search("", ANY_DAAC_KEYWORD, "1", collection_count)
+   search_iterator, count = Cmr.collection_search("", ANY_DAAC_KEYWORD, "1", collection_count)
```
ryanmiller-1: I'm guessing this is a remnant of debugging.

app/helpers/application_helper.rb
```diff
@@ -3,31 +3,68 @@ module ApplicationHelper
...
+  MDQ_PROVIDERS = ['SCIOPS',
+                   'NOAA_NCEI',
+                   'JAXA',
+                   'ISRO',
+                   'AU_AADC',
+                   'ESA',
+                   'EUMETSA',
+                   'MDQTEST']
+
+  def application_mode
```
cgokey: Comments on how this works and all possible modes

jceaser: the only thing that comes to mind is maybe using a state machine in case we change the logic in the future, but I don't think that is an exact fit either. your implementation is fine.

app/models/cmr.rb
```diff
@@ -53,7 +53,11 @@ def self.get_raw_concept(concept_id, revision_id = nil)
   # If Cmr revision id isn't the latest, we pull the latest and ingest it.
 
   def self.update_collections(current_user)
-    records = Record.where state: [Record::STATE_OPEN.to_s, Record::STATE_IN_ARC_REVIEW.to_s, Record::STATE_READY_FOR_DAAC_REVIEW.to_s], recordable_type: "Collection"
+    if current_user.mdq_curator?
+      records = Record.all_records(:mdq_mode).where state: [Record::STATE_OPEN.to_s, Record::STATE_IN_ARC_REVIEW.to_s, Record::STATE_READY_FOR_DAAC_REVIEW.to_s], recordable_type: "Collection"
```
jceaser: what if instead of thinking of this as a user mode, what if it was a filter mode. There could be any number of actions in the future that changes how you're filtered. Maybe the user will say they only want to view one provider at a time or something? Then I am suggesting that when you login your role gives you a default filter rule. Today this may never change, but it could.

db/schema.rb
jceaser: yea, I think this change needs to be rolled back, we should keep the serial

app/models/cmr.rb
cgokey: I agree, it should be the application_mode not the specific mode.   This actually might be a bug, need to look closer.  good catch.

db/schema.rb
cgokey: agree

db/schema.rb
cgokey: I'll do this with regard to testing performance, want to get a base metric first.

db/schema.rb
cgokey: good catch, yes this needs to be removed, was. a change I made a while back in testing that need to be reverted.

app/models/ability.rb
ryanmiller-1: Have you written a ticket or talked to Alicia about this?

cgokey: This should be resolved, as user.mdq_curator? now checks daac as well.

ryanmiller-1: My concern is whether we need to validate (pseudocode) user.daac = <data>.daac

app/helpers/application_helper.rb
charleshuang80: so all daac_curators go into arc_mode? this isn't explained in the comment

app/models/ability.rb
cgokey: Thomas did you end up writing a ticket for this?

app/helpers/application_helper.rb
cgokey: Added a little clearer comments on how the application_mode is determined.

app/models/cmr.rb
cgokey: This isn't a bug the Record.all_records takes an application mode as an argument, it just happens to be we determine the mode is based on the user's role, but we can change the application_mode to whatever we want in the future.

app/helpers/application_helper.rb
ryanmiller-1: This conversation returns me to "I dislike how much we use curator in this application".  

mdq_curator? returns true when either a user has ARC Curator like privileges in non-EOSDIS providers or when their associated daac is in the non-EOSDIS provider list.  In this case, both "Reviewers" and "Reviewees" are MDQ Curators.

ryanmiller-1: I think the case on the ground is that we will have no DAAC level MDQ users at the moment, but I think we will need to come up with a clear way of differentiating the two levels of MDQ curators when we do.

cgokey: well the goal here is to determine the mode... i.e., is it mdq_mode or arc_mode.   not necessarily whether they are a curator or not.

cgokey: so it might be mixing/confusing things by saying the "current_user" is a mdq_curator

cgokey: Wonder if this should be renamed now, mdq_user?

cgokey: I think renamed this actually makes sense and goes more with how it is being used.  Agree?

app/models/cmr.rb
```diff
@@ -53,7 +53,11 @@ def self.get_raw_concept(concept_id, revision_id = nil)
   # If Cmr revision id isn't the latest, we pull the latest and ingest it.
 
   def self.update_collections(current_user)
-    records = Record.where state: [Record::STATE_OPEN.to_s, Record::STATE_IN_ARC_REVIEW.to_s, Record::STATE_READY_FOR_DAAC_REVIEW.to_s], recordable_type: "Collection"
+    if current_user.mdq_curator?
```
ryanmiller-1: This is related to record refreshing, correct?  
Can the MDQ curators access that functionality through the UI?

cgokey: no it is turned off for now.   only admin functionality has that turned on, but we can turn this on for them pretty easily when we want.   This is the conversation I had with Alicia where some of the features where only available to admin's, so we'll need to think about whether we want a "admin" for mdq_curators or not.

app/models/user.rb
charleshuang80: we can ask to change the ACL, and therefore change it here if we want. like we might want to change the ACL name for subscriptions.

PR #153, opened 2020-03-06
Hotfix/CMRARC-545-2 Updating for requested changes and message-id

ryanmiller-1:
The Message-ID does not appear to be sufficient to get gmail delivery, but is necessary, so I left the changes in.  The rest of the changes are to accommodate a request to not show granules.

### DIFF
--- a/app/views/daac_curator_mailer/released_records_digest_notification.html.erb
+++ b/app/views/daac_curator_mailer/released_records_digest_notification.html.erb
@@ -13,7 +13,7 @@
     <tbody>
       <% @records.first(50).each do |record| %>
         <tr>
-          <td><%= link_to(record.collection? ? record.short_name : record.long_name, reports_review_url(record_id: record.id)) %></td>
+          <td><%= link_to(record.short_name, reports_review_url(record_id: record.id)) %></td>
           <td><%= record.released_to_daac_date %></td>
         </tr>
       <% end %>
--- a/lib/rake_helpers/daac_curator_mailer_helper.rb
+++ b/lib/rake_helpers/daac_curator_mailer_helper.rb
@@ -18,7 +18,7 @@ class DaacCuratorMailerOrchestration
     # For each daac where anyone receives an e-mail...
     daacs.each do |daac|
       # Find relevant records or abort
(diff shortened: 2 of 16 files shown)


app/views/daac_curator_mailer/released_records_digest_notification.html.erb
```diff
@@ -13,7 +13,7 @@
     <tbody>
       <% @records.first(50).each do |record| %>
         <tr>
-          <td><%= link_to(record.collection? ? record.short_name : record.long_name, reports_review_url(record_id: record.id)) %></td>
+          <td><%= link_to(record.short_name, reports_review_url(record_id: record.id)) %></td>
```
ryanmiller-1: Don't need this logic to accommodate granules since we will not be displaying them anymore

lib/rake_helpers/daac_curator_mailer_helper.rb
```diff
@@ -18,7 +18,7 @@ def self.released_records_digest_conductor(now: Time.now)
     # For each daac where anyone receives an e-mail...
     daacs.each do |daac|
       # Find relevant records or abort
-      records = Record.where(state: :in_daac_review, daac: daac).order(released_to_daac_date: :desc)
+      records = Record.where(state: :in_daac_review, recordable_type: 'Collection', daac: daac).order(released_to_daac_date: :desc)
```
ryanmiller-1: filter to only display collections

test/mailers/daac_curator_mailer_test.rb
```diff
@@ -5,7 +5,7 @@ class DaacCuratorMailerTest < ActionMailer::TestCase
 
   describe 'released_records_digest_notification email appearance' do
     it 'correctly populates the fields of an email' do
-      email = DaacCuratorMailer.released_records_digest_notification([User.find(7), User.find(8)], [Record.find(22), Record.find(23)], 'OB_DAAC')
+      email = DaacCuratorMailer.released_records_digest_notification([User.find(7), User.find(8)], [Record.find(22), Record.find(25)], 'OB_DAAC')
```
ryanmiller-1: Update test to have two collections instead of a collection and a granule.

PR #159, opened 2020-03-23
CMRARC-572 referenced new omniauth gem.

cgokey:


charleshuang80:
per our conversation with Kevin, and maybe pending result of emails, this can be closed, right?

PR #160, opened 2020-03-25
CMRARC-555 - updates to use jquery3

cgokey:


cgokey:
> Did something complain about the (document) versus ( document )? I don't have a strong preference, just curious

No, it didn't complain, I just copied and pasted from another JS file that had the turbolink added and just pasted that version to all the files.

PR #161, opened 2020-03-26
Feature/CMRARC-377 Fixing filter by campaign

ryanmiller-1:


test/features/campaign_filtering_test.rb
```diff
@@ -0,0 +1,204 @@
...
+Dir[Rails.root.join('test/**/*.rb')].each { |f| require f }
+
+class CanShowCollectionsTest < Capybara::Rails::TestCase
+  include Helpers::UserHelpers
+
+  before do
+    OmniAuth.config.test_mode = true
+  end
+
+  describe 'Campaign Filtering', js: true do
```
ryanmiller-1: remove js: true

app/helpers/application_helper.rb
```diff
@@ -74,11 +74,24 @@ def records_sorted_by_short_name(records)
...
     select_list = [ANY_CAMPAIGN_KEYWORD]
 
-    camps = RecordData.select(:value).where(column_name: "Campaigns/Campaign/ShortName").where.not(value: "").order(:value).distinct
-    select_list.concat(camps.map(&:value))
+    if current_user.daac_curator?
+      daac = current_user.daac
+      state = form == home_path ? :in_daac_review : [:finished, :closed]
+    elsif current_user.mdq_curator?
+      daac = MDQ_PROVIDERS
+      state = form == home_path ? [:open, :in_arc_review, :ready_for_daac_review, :in_daac_review] : [:finished, :closed]
```
ryanmiller-1: check to see if in_daac_review should be removed here

test/features/campaign_filtering_test.rb
```diff
@@ -0,0 +1,203 @@
...
+          it 'can see unfiltered records' do
+            assert has_content?('campaign_test_collection_1-PODAAC')
+            assert has_content?('campaign_test_collection_2-PODAAC')
+            assert has_content?('campaign_test_collection_3-OB_DAAC')
+          end
+        end
+      end
+    end
+
+    describe 'when the user is an MDQ Curator' do
```
ryanmiller-1: check for at least one non-mdq record

test/fixtures/record_data.yml
```diff
@@ -266,3 +266,21 @@ digest_email_collection3_short_name:
   record_id: 25
   column_name: ShortName
   value: Test Short Name3
+
```
ryanmiller-1: these might be superfluous

app/helpers/application_helper.rb
cgokey: I don't think a mdc_curator? can see in_daac_review state, but you were going to double check.

app/helpers/application_helper.rb
```diff
@@ -74,11 +74,24 @@ def records_sorted_by_short_name(records)
...
+    elsif current_user.arc_curator?
+      daac = ARC_PROVIDERS
+      state = form == home_path ? [:open, :in_arc_review, :ready_for_daac_review] : [:finished, :closed]
```
cgokey: there is no in_daac_review here, so changing above would be consistent.

app/helpers/record_helper.rb
```diff
@@ -1,4 +1,10 @@
 module RecordHelper
+  # Campaigns/Campaign/ShortName comes from ECHO10
+  # Projects/ShortName comes from UMM-JSON
+  # Project/Short_Name comes from DIF10
+  CAMPAIGN_COLUMNS = %w[Campaigns/Campaign/ShortName
+                        Projects/ShortName
+                        Project/Short_Name]
```
cgokey: look in record_formats/dif10_record.rb, record_formats/echo10_record.rb, record_formats/umm_record.rb, I feel like the logic for pulling out the right column name for campaign should be in there and you can just say:
record.campaign_column
This has the benefit if a new format is added, it will be apparent what needs to be implemented for that new format.

app/models/metadata.rb
```diff
@@ -21,4 +21,9 @@ def get_records(descending = true)
   def self.daac_from_concept_id(concept_id)
     concept_id.partition('-').last
   end
+
+  def self.extract_campaign(collection_data)
+    keyword = RecordHelper::CAMPAIGN_COLUMNS.select { |column| collection_data.keys.include?(column)}
```
cgokey: hmm, my idea might not working because you are extracting this at the metadata level and not at the record level but worth thinking about if it is possible to use that record_formats if possible.

app/helpers/application_helper.rb
ryanmiller-1: I've made the change, but I haven't pushed it because I was waiting for other feedback.

app/helpers/record_helper.rb
ryanmiller-1: Doing it that way means inspecting every record after the DB query.  Doing it this way means the DB does that work for us.  I would prefer not to change it, but I should find some place to put comments to try to make this more maintainable.

app/models/collection.rb
```diff
@@ -37,7 +37,7 @@ def self.create_record(concept_id, revision_id, data_format, collection_data, op
     Collection.transaction do
       collection = Collection.find_or_create_by(concept_id: concept_id)
       collection.update_attributes!(short_name: short_name)
-      new_record = Record.create(recordable: collection, revision_id: revision_id, format: data_format, daac: daac_from_concept_id(concept_id))
+      new_record = Record.create(recordable: collection, revision_id: revision_id, format: data_format, daac: daac_from_concept_id(concept_id), campaign: extract_campaign(collection_data))
```
cgokey: So you could do it here.   Can you override Record.create?  don't pass it a campaign, do something like this (pseudocode)
def create(... args ...) 
  record = super.create(... args ..)
  # record is not a specific record format instance.
  record.campaign = record[campaign_column]
  return record
end

charleshuang80:
I have no comments at the moment (asked Ryan some questions directly). But would like the see the changes that are going to be pushed up.

app/helpers/record_helper.rb
```diff
@@ -75,4 +81,13 @@ def should_show_copy_recommendations_button?(record)
...
     record.copy_recommendations_note.nil? ? 'eui-btn--green' : 'eui-btn--disabled'
   end
+
+  # When a record has multiples, Dashboard stores them in the value field
+  # formatted like: "• <value1>\n• <value2>\n ..."
+  # Returns an array like: [value1, value2, ...]
+  def clean_up_campaign(campaign)
+    return [] if campaign.nil?
+
+    campaign.tr('•', '').split("\n").each(&:strip!)
```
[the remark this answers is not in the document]
ryanmiller-1: `["Aqua", "LANDSAT-7", "", "TERRA", "Terra", "• Terra\n• Aqua", "OV-105", "SUOMI-NPP", "• GLONASS\n• GPS", "• Beidou\n• Galileo\n• GLONASS\n• GPS\n• GROUND STATIONS\n• QZSS\n• IRNSS", "• Aqua\n• Terra", "In Situ Ocean-based Platforms", "Suomi-NPP", "STS-99", "NOAA-20", "ISS", "• Terra\n• NOAA-7\n• NOAA-9\n• NOAA-11\n• NOAA-14", "• Terra\n• NOAA-7\n• NOAA-9\n• NOAA-11\n• NOAA-14\n• LANDSAT-5\n• LANDSAT-7", "• LANDSAT-7\n• DEM\n• LANDSAT-5", "• LANDSAT-8\n• SENTINEL-2\n• DEM\n• LANDSAT-5", "• LANDSAT-8\n• LANDSAT-7\n• DEM\n• LANDSAT-5", "• LANDSAT-8\n• LANDSAT-7\n• DEM", "• LANDSAT-8\n• LANDSAT-7\n• LANDSAT-5\n• SENTINEL-2", "• LANDSAT-7\n• LANDSAT-5\n• Terra", "• NOAA-7\n• NOAA-9\n• NOAA-11\n• NOAA-14\n• Terra", "EO-1", "• LANDSAT-7\n• LANDSAT-5", "Sentinel-3A", "Sentinel-3B", "G-IV", "AC-680E"]`
That is a list of platform short names being stored in dashboard.  Each string is displayed as a list in the review section.  Campaigns (and every other multi item I've looked at, I chose platforms at a whim here) are stored the same way.  
I am not removing it from the column which is being used to display data.  I am adding a new column (to a different table) and reformatting the data so that it is more programmatically usable.  Digging deeper into it, the bullet is being added as part of ingest and is character `bullet = "\u{2022} "`.

app/helpers/application_helper.rb
```diff
@@ -74,11 +74,24 @@ def records_sorted_by_short_name(records)
...
     select_list = [ANY_CAMPAIGN_KEYWORD]
 
-    camps = RecordData.select(:value).where(column_name: "Campaigns/Campaign/ShortName").where.not(value: "").order(:value).distinct
-    select_list.concat(camps.map(&:value))
+    if current_user.daac_curator?
+      daac = current_user.daac
+      state = form == home_path ? :in_daac_review : [:finished, :closed]
+    elsif current_user.mdq_curator?
+      daac = MDQ_PROVIDERS
+      state = form == home_path ? [:open, :in_arc_review, :ready_for_daac_review] : [:finished, :closed]
```
charleshuang80: not anything to be addressed now, but wondering if there should be a `:in_mdq_review` state?

cgokey: I was thinking about that too, but it might just want to be :in_review.  or :in_curator_view

ryanmiller-1: Have we written any tickets based on what MDQ curators might want yet?

PR #162, opened 2020-03-26
CMRARC-518 - Ensure granule and associate granule stay in sync

cgokey:


app/models/record.rb
```diff
@@ -28,11 +28,22 @@ class Record < ApplicationRecord
...
   scope :metadata_format, ->(format) { where(format: format) }
 
+  after_update :update_associated_granule_state
+
+  def update_associated_granule_state
+    unless previous_changes['state'].nil?
+      if recordable_type == 'Collection' && has_associated_granule?(self)
+        granule_record = Record.find_by id: self.associated_granule_value
+        granule_record.state = self.state
+        granule_record.save
```
cgokey: maybe say: granule_record.update(state: self.state) instead

app/controllers/records_controller.rb
```diff
@@ -193,13 +198,41 @@ def find_record
     redirect_to home_path unless @record
   end
 
+  def granule_valid?(granule_record, collection_state, associating_flag)
```
cgokey: Maybe should go in the helper class actually.

test/controllers/records_controller_test.rb
```diff
@@ -154,6 +155,69 @@ class RecordsControllerTest < ActionController::TestCase
       assert_equal 'The record C1000000020-LANCEAMSR2 was successfully updated.', flash[:notice]
       assert_equal Record.find(18).state, 'ready_for_daac_review'
     end
+
+    it 'Mark complete checks on BOTH the collection review AND the associated granule review.' do
```
cgokey: fix sentence.

test/controllers/records_controller_test.rb
```diff
@@ -188,6 +252,7 @@ class RecordsControllerTest < ActionController::TestCase
       assert_equal Record.find(12).state, 'in_daac_review'
       assert_equal Record.find(17).state, 'in_daac_review'
       post :associate_granule_to_collection, params: { id: 12, associated_granule_value: 17 }
+      puts "flash=#{flash['alert']}"
```
cgokey: remove

app/controllers/records_controller.rb
```diff
@@ -223,10 +256,7 @@ def completion_success(record)
   def release_record_for_daac_review(record)
     success = false
     Record.transaction do
-      case record.recordable_type
-      when 'Granule'
-        success = release_granule_record_to_daacs(record)
-      when 'Collection'
+      if record.recordable_type == 'Collection'
```
ryanmiller-1: record.collection?

test/controllers/records_controller_test.rb
```diff
@@ -146,6 +146,7 @@ class RecordsControllerTest < ActionController::TestCase
       assert_nil Record.find(18).released_to_daac_date
       assert_equal Record.find(18).state, 'ready_for_daac_review'
       post :complete, params: { id: 18 }
+      puts "flash=#{flash[:alert]}"
```
ryanmiller-1: rm

test/controllers/records_controller_test.rb
ryanmiller-1: These tests can be condensed a bit.

test/controllers/records_controller_test.rb
```diff
@@ -279,4 +344,7 @@ class RecordsControllerTest < ActionController::TestCase
       assert_equal 'Deleted the following collections: C1000000020-LANCEAMSR2/8 C1000000020-LANCEAMSR2/9 ', flash[:notice]
     end
   end
+
```
ryanmiller-1: rm this and the white space in associating_granules_test

test/features/associating_granules_test.rb
```diff
@@ -47,7 +48,7 @@ class AssociatingGranulesTest < Capybara::Rails::TestCase
 
     end
 
-    describe 'associated granule reports' do
+    describe 'associated granule reports', js: true do
```
ryanmiller-1: debugging or required?

app/views/reviews/_navbar.html.erb
```diff
@@ -135,9 +135,11 @@
         };
       </script>
 
-      <%= button_to complete_record_path, disabled: disable_complete_button?(@reviews, @record), id: "done_button", class: "eui-btn eui-btn--round eui-btn--sm mark_complete_button" do %>
-        <p class="button_text"><%= complete_button_text(@record) %></p>
-        <i class="eui-icon eui-fa-link button_symbol complete_symbol"></i>
+      <% if @record.recordable_type == 'Collection' %>
```
ryanmiller-1: @record.collection?

app/models/record.rb
ryanmiller-1: Isn't this going to skip the AASM validations?

test/features/perform_sanity_checks_on_reviews_test.rb
```diff
@@ -0,0 +1,194 @@
...
+
+  before do
+    stub_request(:get, "#{Cmr.get_cmr_base_url}/search/granules.echo10?concept_id=G309210-GHRC")
+      .with(
+        headers: {'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Accept' => '*/*', 'User-Agent' => 'Ruby'}
+      )
+      .to_return(status: 200, body: '<?xml version="1.0" encoding="UTF-8"?><results><hits>0</hits><took>32</took></results>', headers: {})
+  end
+
+  describe 'performs "mark complete" on a collection record' do
```
ryanmiller-1: I really like the idea of more feature tests for dashboard.  It's a huge help to people coming onto a new app to have some baseline of 'did I break it?'

test/features/perform_sanity_checks_on_reviews_test.rb
```diff
@@ -0,0 +1,194 @@
...
+    before do
+      OmniAuth.config.test_mode = true
+      mock_login(role: 'arc_curator')
+    end
+
+    # in this case short name, long name are marked with a color, but version id is not
+    it 'will perform sanity checks on BOTH collection and assoc granule review' do
```
ryanmiller-1: I am concerned it might be difficult to debug this test when the workflow breaks.  Breaking it into smaller tests may make it more digestible.

cgokey: I agree, there is no helpers for tests, feel like this should be broken up into a few helpers as well.

app/models/record.rb
cgokey: Yes, but only for the granule side.   The validations are now done by calling the methods directly on the record model object during association or changing the state of the collection.  This also provides slightly better and more specific error messages since the validation is being done on two different records at the same time.

app/helpers/record_helper.rb
```diff
@@ -4,6 +4,31 @@ def is_number?(object)
     true if Float(object) rescue false
   end
 
+  def has_associated_granule?(record)
+    record.associated_granule_value.present? && is_number?(record.associated_granule_value) ? true : false
```
ryanmiller-1: record.associated_granule_value.present? && is_number?(record.associated_granule_value) will evaluate to true or false, won't it?

test/controllers/records_controller_test.rb
```diff
@@ -154,6 +154,69 @@ class RecordsControllerTest < ActionController::TestCase
       assert_equal 'The record C1000000020-LANCEAMSR2 was successfully updated.', flash[:notice]
       assert_equal Record.find(18).state, 'ready_for_daac_review'
     end
+
```
ryanmiller-1:
```
      user = User.find_by(email: 'arc_curator@element84.com')
      sign_in(user)
      stub_urs_access(user.uid, user.access_token, user.refresh_token)
      record = Record.find(1)
      post :associate_granule_to_collection, params: { id: record.id, associated_granule_value: 16 }
      assert_equal 'Granule G309210-GHRC/6 has been successfully associated to this collection revision 8. ', flash[:notice]
```
Can be moved to a shared before block.

test/features/perform_sanity_checks_on_reviews_test.rb
```diff
@@ -0,0 +1,166 @@
...
+      assert_equal Record.find_by(id: 1).state, 'in_arc_review'
+      assert_equal Record.find_by(id: 1).state, Record.find_by(id: 16).state
+
+      assert has_no_content? 'Not all columns in the associated granule have been flagged with a color!'
+      assert has_no_content? 'The associated granule needs two completed reviews'
+      assert has_content? 'Not all columns have been flagged with a color, cannot close review.'
+      assert has_content? 'Record failed to update.'
+    end
+
+    it 'fixes granule, fixes collection and produces success message -- should ignore second opinion checks, then fixes second opinions checks' do
```
ryanmiller-1: This test is testing three different bits of related functionality.  In the MMT, we would typically do this with nested context blocks.  
```
context
  before do
    the stuff
  end

  it do
    test stuff
  end

  context
    before do
      some more stuff
    end
    ...
```
And I think it would make this test more maintainable.

test/features/perform_sanity_checks_on_reviews_test.rb
```diff
@@ -0,0 +1,166 @@
...
+      find('#color_button_green_VersionId').click
+      click_button 'Save Review'
+      find('#nav_back_button').click
+
+      # produces success message
+      click_button 'MARK AS DONE'
+      assert_equal Record.find_by(id: 1).state, 'ready_for_daac_review'
+      assert_equal Record.find_by(id: 1).state, Record.find_by(id: 16).state
+      assert has_content? 'Record has been successfully updated.'
+
```
ryanmiller-1: If I were breaking this into tests, what is above is the first context.

test/features/perform_sanity_checks_on_reviews_test.rb
```diff
@@ -0,0 +1,166 @@
...
+      # Now lets try to release to daac, it should fail because collection should fail the second opinion check,
+      # which is only checked when being released to a daac
+      see_collection_revision_details(8)
+
+      # Now the button should say "RELEASE TO DAAC" since we are in ready_for_daac_review state
+      click_button 'RELEASE TO DAAC'
+      assert has_content? 'Record failed to update.'
+      assert has_content? 'Some columns still need a second opinion review, cannot close review.'
+
```
ryanmiller-1: Second above

test/controllers/records_controller_test.rb
cgokey: Take a look at the latest nested workflow and tell me if this comment you made still makes sense.  not sure exactly what you mean with the latest.

test/features/perform_sanity_checks_on_reviews_test.rb
cgokey: likewise with this one.

app/helpers/record_helper.rb
```diff
@@ -4,6 +4,31 @@ def is_number?(object)
     true if Float(object) rescue false
   end
 
+  def has_associated_granule?(record)
+    record.associated_granule_value.present? && is_number?(record.associated_granule_value)
+  end
+
+  def granule_valid?(granule_record, collection_state, associating_flag)
```
charleshuang80: a little less sure about this, but I would also ask if it makes sense to make this an instance method for either granules or records

test/features/perform_sanity_checks_on_reviews_test.rb
```diff
@@ -0,0 +1,150 @@
+require 'test_helper'
+Dir[Rails.root.join('test/**/*.rb')].each { |f| require f }
+
+class PerformsSanityChecksOnReviewsTest < Capybara::Rails::TestCase
```
charleshuang80: I also like that you're adding these tests. I think the comments are quite helpful too.

app/controllers/records_controller.rb
```diff
@@ -37,9 +37,17 @@ def associate_granule_to_collection
     else
       granule_record = Record.find_by id: associated_granule_value
       if !granule_record.nil?
```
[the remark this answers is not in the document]
jceaser: it's the ruby thing to do right

app/helpers/record_helper.rb
```diff
@@ -4,6 +4,31 @@ def is_number?(object)
     true if Float(object) rescue false
   end
 
+  def has_associated_granule?(record)
```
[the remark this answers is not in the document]
cgokey: agree good catch

app/helpers/record_helper.rb
cgokey: I think there is just a little too much business logic in this method to make it part of the model object.  prefer to keep it as a helper function on this one.

jceaser: what makes it valid? as in, can you add something descriptive to the name, one more word?

test/features/associating_granules_test.rb
```diff
@@ -73,6 +74,8 @@ class AssociatingGranulesTest < Capybara::Rails::TestCase
         page.must_have_content('RECORD METRICS') # verify it has report metrics
         page.assert_selector('.checked_num', count: 2) # of elements reviewed appears twice.
       end
+
+
```
jceaser: I guess you have to remove this too

app/helpers/record_helper.rb
cgokey: there is a few different things going on in there, hard to wrap into one word.  Suggestion?

charleshuang80: to Thomas' questions, maybe a comment could also suffice.

I'll defer to your judgement in terms of keeping the method here, so no need to move. But two thoughts:
- it seems a lot of this is asking things of the granule record, that could be gotten in the model
- 'they' often say that you should put business logic in models instead of controllers, so I think it would make sense to have business logic there

-

cgokey: I can definitely add a comment.   But what strikes me wrong about moving it to the model is that this is very specific to granules, and the model object is a Record object which could be a granule or collection.  I guess by definition the helper is included in the controller, so could be seen as bloating the controller.   But I consider it more of a helper function to help the application logic determine what to do.

cgokey: Thomas I renamed it to associated_granule_valid? and added comments to all that it is doing.

cgokey: I opted to split this helper function into two different functions and get rid of the flag variable.   I think it makes things a bit more understandable.

PR #163, opened 2020-03-31
CMRARC-505 - now checks http response code.

cgokey:


app/models/cmr.rb
```diff
@@ -369,8 +369,10 @@ def self.get_raw_granule_results(concept_id)
   # returns true if a granule with the specified concept id exists in cmr
   def self.raw_granule?(concept_id)
     url = Cmr.api_url('granules', 'echo10', 'concept_id': concept_id)
-    granule_xml = Cmr.cmr_request(url).parsed_response
-    granule_results = convert_to_hash('echo10', granule_xml)['results']
+    response = Cmr.cmr_request(url)
+    return false if response.code != 200
```
charleshuang80: would it be helpful to log anything here?

app/models/cmr.rb
```diff
@@ -686,6 +688,7 @@ def self.get_cmr_base_url
     if cmr_base_url.nil?
       cmr_base_url = 'https://cmr.earthdata.nasa.gov'
     end
+    cmr_base_url = 'https://cmr.earthdata.nasa.gov'
```
charleshuang80: this either seems redundant or that you can remove all the lines above it in the method.

app/models/cmr.rb
cgokey: I added a log message.

app/models/cmr.rb
cgokey: debugging accidently made it into the commit.  removed.

PR #168, opened 2020-05-07
cmrarc-579 Add status 'closed'

htranho:


app/helpers/record_helper.rb
```diff
@@ -7,7 +7,7 @@ def is_number?(object)
   # note if the checks fail, the caller should not associate the granule record to the collection.
   def can_associate_granule?(granule_record, collection_state)
     # it is ok to associate granule records without checks if in open or in_arc_review
-    return [true, nil] if %w(open in_arc_review).include?(collection_state)
+    return [true, nil] if %w(open in_arc_review closed).include?(collection_state)
```
cgokey, line 10: Looks good, but make a comment above it that this is a temporary change to allow users to associate closed records until arc team either finishes reviewing all legacy imported records or we fix the bug that is preventing users from marking all colors for a record.

PR #177, opened 2020-07-01
CMRARC-534 Copy over recommendations from a different concept ID

htranho:


htranho:
need to redo, code mixed with other branch because rebase from master

PR #189, opened 2020-09-04
CMRARC-592 combine daacs for filtering, search, reporting.

cgokey:


db/migrate/20200904121332_combine_providers.rb
```diff
@@ -0,0 +1,44 @@
+class CombineProviders < ActiveRecord::Migration[5.2]
+  def up
+    Record.all.visible.where(daac: ["LARC_ASDC", "LARC"]).find_each do |record|
```
cgokey, line 3: remove .all.visible

htranho (approved):

PR #203, opened 2020-09-24
Feature/cmrarc 601 Reduce size of color select box

htranho:


htranho:
Mixed up with other changes

PR #223, opened 2020-12-14
CMRARC-630 - added login entry for use by google analytics.

cgokey:


app/controllers/login_controller.rb
```diff
@@ -7,6 +7,7 @@ def urs
 
       instance = UserSingleton.instance
       instance.current_user = @user
+      Rails.logger.info("Google Analytics - Users #{@user.uid} from provider/role #{@user.role}")
```
cgokey, line 10: How often do users need to login

htranho (approved):

PR #225, opened 2021-02-04
CMRARC-656 - increased test coverage to bump it up over 80%

cgokey:


app/models/cmr.rb
```diff
@@ -582,11 +582,13 @@ def self.json_collection_search(free_text, provider = ANY_DAAC_KEYWORD, curr_pag
 
       begin
         raw_json = Cmr.cmr_request(query_text).parsed_response
```
cgokey, line 584: maybe remove parsed_response to be safe.

htranho (approved):

PR #231, opened 2021-02-24
CMRARC-645 added db layer needed for invalid keywords.

cgokey:


db/schema.rb
```diff
@@ -48,6 +48,20 @@
     t.index ["user_id"], name: "index_ingests_on_user_id"
   end
 
+  create_table "invalid_keywords", force: :cascade do |t|
+    t.datetime "last_updated"
```
cgokey, line 52: remove this.

PR #249, opened 2021-03-11
CMRARC-650 changes for dif10

cgokey:


app/models/record_formats/dif10_record.rb
```diff
@@ -96,7 +96,7 @@ def long_name_field
     end
 
     def controlled_element_map
-      CONTROLLED_ELEMENT_MAP
+      ControlledElements.instance.mapping("umm_json")
```
cgokey, line 99: needs dif10 not umm_json!

htranho (approved):

PR #263, opened 2021-03-23
Cmrarc 640 Update required field list

htranho:


app/models/umm_desired_fields.rb
```diff
@@ -0,0 +1,16 @@
+class UmmDesiredFields
```
htranho, line 1: remove umm in names of 3 classes, format as input param: ummC,dif10,echo10

cgokey (approved):

PR #272, opened 2021-03-30
CMRARC-688 Update for dif10

htranho:


data/dif10_desired_fields.csv
```diff
@@ -58,6 +58,11 @@ Organization/Organization_Name/Long_Name
 Organization/Organization_URL
 Organization/Personnel/Contact_Group/Name
 Distribution/Distribution_Format
+DirectDistributionInformation/Region
+DirectDistributionInformation/S3BucketAndObjectPrefixName
+DirectDistributionInformation/S3CredentialsAPIEndpoint
+DirectDistributionInformation/S3CredentialsAPIDocumentationURL
+Use_Constraints
```
htranho, line 65: remove parent field

data/dif10_section_titles.csv
```diff
@@ -2,7 +2,6 @@ dif10_section_title
 Summary
```
htranho, line 2: All top level field which has children, check if need to add field as simple array

cgokey (approved):

PR #273, opened 2021-03-31
CMRARC-687 Update field list for echo10

htranho:


data/echo10_section_titles.csv
```diff
@@ -1,18 +1,25 @@
 echo10_section_title
 DOI
+AssociatedDOIs
+UseConstraints
+TemporalKeywords
```
htranho, line 5: Remove simple list

cgokey (approved):

PR #307, opened 2021-07-16
CMRARC-730: successfully retrieved CMR collection using user-supplied parameters.

hchoee:
I created the controller that will house the bulk of the code for my project, named gkr_keyword_comparison_controller. I added 2 routes, one for rendering the page where the user can supply a concept ID and threshold for their desired collection, and another for grabbing the parameters from the user. The concept ID is then used to retrieve the collection from CMR, using the umm_json format. The format I supplied to the call to CMR was based off of other calls I found in dashboard.


app/assets/javascripts/gkr_keyword_comparison.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
```
ChristianTrummer99, line 3: This file can be deleted

app/assets/stylesheets/gkr_keyword_comparison.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the GkrKeywordComparison controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: https://sass-lang.com/
```
ChristianTrummer99, line 3: This file can be deleted

config/routes.rb
```diff
@@ -1,4 +1,6 @@
 Rails.application.routes.draw do
+  get 'gkr_keyword_comparison/show'
+  get 'gkr/show'
```
ChristianTrummer99, line 3: Are these necessary? ^

test/controllers/gkr_keyword_comparison_controller_test.rb
```diff
@@ -0,0 +1,9 @@
+require 'test_helper'
+
+class GkrKeywordComparisonControllerTest < ActionDispatch::IntegrationTest
+  test "should get show" do
+    get gkr_keyword_comparison_show_url
+    assert_response :success
+  end
```
ChristianTrummer99, line 7: Instead of `get gkr_keyword_comparison_show_url` I would use `visit gkr_keyword_comparison_path` and verify that the form you created is being rendered properly. Additionally, you'll want to test that you can retrieve collections from CMR successfully. So maybe once your on the show page you can simulate someone filling out the form and then asserting that the proper collection is coming back from cmr

app/helpers/gkr_keyword_comparison_helper.rb
```diff
@@ -0,0 +1,2 @@
+module GkrKeywordComparisonHelper
+end
```
ChristianTrummer99, line 2: This file can be deleted

app/views/gkr_keyword_comparison/new.html.erb
```diff
@@ -0,0 +1,11 @@
...
+    <%= label_tag :concept_id %><br>
+    <%= text_field_tag :Concept_ID %>
+    <br>
+    <br>
+    <%= label_tag :threshold %><br>
+    <%= text_field_tag :Threshold %>
+    <p>
+        <%= submit_tag("Submit") %>
+    </p>
+<% end %>
```
ChristianTrummer99, line 11: I don't think you need this file either, the `show` page should be sufficient

ChristianTrummer99 (changes requested):

app/views/gkr_keyword_comparison/new.html.erb
hchoee: You're right, I believe I just forgot to delete that after I was trying to troubleshoot something

config/routes.rb
hchoee: No, I think they might have automatically been added when I created new views. I don't remember putting them there.

app/controllers/gkr_keyword_comparison_controller.rb
```diff
@@ -0,0 +1,10 @@
+class GkrKeywordComparisonController < ApplicationController
+  def show
+  end
+
+  def create
+    concept_id = params[:concept_id]
+    threshold = params[:threshold]
+    collection_hash = Cmr.get_collection(concept_id, 'umm_json') # using this method bc get_concept() did not work for all concept ids
```
cgokey, line 8: Do you want to make collection_hash available to the view, if so you can put a @ in front of it.  Maybe also just rename it to @collection

test/controllers/gkr_keyword_comparison_controller_test.rb
cgokey: Agree, so you'll want to create a erb that will be render the collection you pull from CMR, just for the purpose of testing, maybe you can add the abstract and keywords to it.   Then you can setup a test that verifies it is all working.

app/controllers/gkr_keyword_comparison_controller.rb
cgokey: Then setup a page that renders the abstract and science keywords maybe.

ChristianTrummer99 (approved):

jdf-eos (approved):

PR #309, opened 2021-07-29
CMRARC-731: Keywords and Abstract successfully pulled from collection

hchoee:
I was able to easily extract the abstract and science keyword values from the collection, which will be needed later on to (1) retrieve recommendations from GKR and (2) compare the keywords from CMR to GKR's recommendations. I then display those on the application. Since I am redirecting to the same view that was rendered originally, I encapsulated the code for displaying the abstract and keywords in if statements to prevent any issues with nil variables.

### DIFF
--- a/app/controllers/gkr_keyword_comparison_controller.rb
+++ b/app/controllers/gkr_keyword_comparison_controller.rb
@@ -8,9 +8,9 @@ class GkrKeywordComparisonController < ApplicationController
     concept_id = params[:concept_id]
     threshold = params[:threshold]
     collection = Cmr.get_raw_collection(concept_id, 'umm_json')
-    abstract_p = collection['Abstract']
-    cmr_keywords_p = collection['ScienceKeywords']
+    abstract_for_gkr = collection['Abstract']
+    curated_science_keywords = collection['ScienceKeywords']
 
-    redirect_to gkr_keyword_comparison_path(abstract: abstract_p, cmr_keywords: cmr_keywords_p)
+    redirect_to gkr_keyword_comparison_path(abstract: abstract_for_gkr, cmr_keywords: curated_science_keywords)
   end
 end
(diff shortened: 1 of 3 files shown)


app/controllers/gkr_keyword_comparison_controller.rb
```diff
@@ -1,14 +1,16 @@
...
   def show
     @abstract = params[:abstract]
+    @cmr_keywords = params[:cmr_keywords]
   end
 
   def create
     concept_id = params[:concept_id]
     threshold = params[:threshold]
     collection = Cmr.get_raw_collection(concept_id, 'umm_json')
     abstract_p = collection['Abstract']
```
ChristianTrummer99, line 11: why the `_p` for these?

app/views/gkr_keyword_comparison/show.html.erb
```diff
@@ -13,5 +13,18 @@
 <% end %>
 <br>
 <br>
-<h1 style='color: white'> Abstract: </h1>
-<p style='color: white'> <%= @abstract %> </p>
\ No newline at end of file
+
+<% if @abstract != nil then %>
```
ChristianTrummer99, line 17: you can remove `!= nil then` and your code will function the same. It is rails convention to only use `then` for inline ternaries so you'll want to remove `then`. Also, if you want to do the `nil` check you can (though I personally don't like it) you should use `unless @abstract.nil?` instead of `!= nil` as it's rails best practice to use the rails built-ins

app/views/gkr_keyword_comparison/show.html.erb
```diff
@@ -13,5 +13,18 @@
...
+  <h1 style='color: white'> Abstract: </h1>
+  <p style='color: white'> <%= @abstract %> </p>
+<% end %>
+
+<% if @cmr_keywords != nil then %>
```
ChristianTrummer99, line 22: Same here

ChristianTrummer99 (approved):

rosycordova (approved):

PR #315, opened 2021-08-19
CMRARC-744: added n/a as default when there is no pyquarc check.

cgokey:


ChristianTrummer99 (approved):

app/views/reviews/show.html.erb
```diff
@@ -98,25 +98,11 @@
         </tr>
         <tr>
```
jdf-eos, line 99: It is difficult to count them in. a diff, but it looks like this <tr> isn't closed with a matching </tr>

jdf-eos (approved):

PR #356, opened 2022-05-07
CMRARC-763: Will now show umm-c version

cgokey:
Also updated to use rails integration tests for the controller controller tests (other way of doing it is being deprecated)

### DIFF
--- a/lib/dashboard_checker.sh
+++ b/lib/dashboard_checker.sh
@@ -2,5 +2,5 @@
 #if [ -d "/opt/rh/rh-python38" ]; then
 #  source /opt/rh/rh-python38/enable
 #fi
-pyenv local 3.10.3
+#pyenv local 3.10.3
 python3 -W ignore lib/dashboard_checker.py $1 $2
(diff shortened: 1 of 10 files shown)


lib/dashboard_checker.sh
```diff
@@ -2,5 +2,5 @@
 #if [ -d "/opt/rh/rh-python38" ]; then
 #  source /opt/rh/rh-python38/enable
 #fi
-#pyenv local 3.10.3
+pyenv local 3.10.3
```
cgokey, line 5: revert

lib/granule_checker.sh
```diff
@@ -2,6 +2,6 @@
 #if [ -d "/opt/rh/rh-python38" ]; then
 #  source /opt/rh/rh-python38/enable
 #fi
-#pyenv local 2.7.18
+pyenv local 2.7.18
```
cgokey, line 5: revert accidenty committed this.

rosycordova (approved):

PR #380, opened 2022-11-30
Cmrarc 789 - logic to call and process pyquarc api

cgokey:


lib/file.txt
```diff
@@ -0,0 +1 @@
+{"name":"hello"}
```
tle1989, line 1: This file was created only for testing purpose right?

lib/dashboard_checker.py
```diff
@@ -5,7 +5,7 @@
 import sys
 import json
 
-# Uses pyQuARC 1.1.5
+# Uses pyQuARC 2.0
```
tle1989, line 8: Maybe 2.0.0 just to have the standard format for future reference.

tle1989 (approved):

lib/file.txt
cgokey: Good catch I'll remove this.

lib/dashboard_checker.py
cgokey: Yeah, it is not that important for code, but it is for bamboo, but hopefully next PI we'll do away with the command line altogether.

cgokey: Actually I'm going to try and remove these files as I don't think they are needed anymore

dmistry1 (approved):

htranho (approved):

rosycordova (approved):

PR #402, opened 2023-04-05
CMRARC-797: Daac user should see only in_daac_review revisions

htranho:


app/controllers/collections_controller.rb
```diff
@@ -14,9 +14,11 @@ def show
     end
     if @record
       @collection = get_collection_from_record(@record)
-
       @concept_id = @collection.concept_id
       @collection_records = @collection.get_records
+      if current_user.daac_curator?
+        @collection_records.select! {|record| record.state == Record::STATE_IN_DAAC_REVIEW.to_s}
+      end
       @granule_objects = Granule.where(collection: @collection)
```
cgokey, line 22: I wonder if we need to do the same thing with granule revisions?

htranho: done

cgokey (approved):

dmistry1 (approved):

PR #416, opened 2023-09-01
CMRARC-783: Fix Import new granule revision

htranho:


app/models/granule.rb
```diff
@@ -101,7 +104,7 @@ def self.create_granule(granule, current_user, granule_info)
       granule_record_data_list.push(granule_record_data)
     end
     granule_record.campaign = ApplicationController.helpers.clean_up_campaign(granule_record.campaign_from_record_data)
-    granule_record.save
```
htranho, line 104: This line produced an empty review, so it is removed. Saving will be done in the caller.

app/controllers/granules_controller.rb
```diff
@@ -60,7 +60,17 @@ def pull_latest
     granule = Granule.find(params[:id])
     collection = granule.collection
     begin
-      Granule.add_new_revision_to_granule(collection.concept_id, granule, current_user)
+      granule_components = Granule.add_new_revision_to_granule(collection.concept_id, granule, current_user)
```
htranho, line 63: This code is copied from the business case 'Add new random granule'  (See add_granule_with_concept_id in models/collection.rb)

app/models/granule.rb
```diff
@@ -101,7 +104,7 @@ def self.create_granule(granule, current_user, granule_info)
...
+    #granule_record.save
```
htranho, line 107: This line produced an empty review, so it is removed. Saving will be done in the caller.

tle1989 (approved):

dmistry1 (approved):

PR #457, opened 2024-04-19
CMRARC-384: Open new tab when clicking ‘UMM-JSON’

mandyparson:
When reviewing a file, clicking UMM_JSON will open a new tab. Added target='_blank' to <a> tag. 

Screenshot 2024-04-19 at 9 37 45 AM

### DIFF
--- a/app/views/reviews/_navbar.html.erb
+++ b/app/views/reviews/_navbar.html.erb
@@ -58,7 +58,7 @@
       <% end %>
 
       <p class="format_links">SOURCE FILES :
-        <a class="format_link" href="<%= record.umm_json_link %>" download>UMM-JSON</a> |
+        <a class="format_link" href="<%= record.umm_json_link %>" download target="_blank">UMM-JSON</a> |
         <a class="format_link" href="<%= record.native_link %>" download>NATIVE</a></p>
     <% else %>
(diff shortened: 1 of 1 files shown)


app/views/reviews/_navbar.html.erb
```diff
@@ -58,7 +58,7 @@
       <% end %>
 
       <p class="format_links">SOURCE FILES :
-        <a class="format_link" href="<%= record.umm_json_link %>" download>UMM-JSON</a> |
+        <a class="format_link" href="<%= record.umm_json_link %>" download target="_blank">UMM-JSON</a> |
         <a class="format_link" href="<%= record.native_link %>" download>NATIVE</a></p>
```
cgokey, line 62: should we open native as well in a new tab?

mandyparson: I'm actually not entirely sure what the link is supposed to do. Currently it opens a dialog box that asks me to name a .native file for download. Is that the expected outcome?

cgokey: I guess so, maybe just keep it the way it is.   The browser probably doesn't know how to handle .native so it is causing a download.

cgokey (approved):

william-valencia (approved):

PR #458, opened 2024-04-19
CMRARC-841:Update CMR Dashboard Keyword Validation Report to include metadata from new CMR Providers

tle1989:
New Providers are added to  provider list for creating the Invalid Keyword Report
LARC_CLOUD
LM_FIRMS
LPCLOUD
GESDISCCLD

Before: 
CMRDASHBOARD SIT

After: 
CMR DASHBOARD LOCAL

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -22,10 +22,7 @@ module ApplicationHelper
                    'PODAAC',
                    'OMINRT',
                    'ARCTEST',
-                   'LARC_CLOUD',
-                   'LM_FIRMS',
-                   'LPCLOUD',
-                   'GESDISCCLD']
+                   'LM_FIRMS']
 
   #providers are specified to identify only the records within EOSDIS which MDQ team curates
   MDQ_PROVIDERS = ['SCIOPS',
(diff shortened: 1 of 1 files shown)


mandyparson:
Couple general questions: 

1. What is the difference between providers and virtual providers? I see that we are treating LM_FIRMS like a provider and LARC_CLOUD, LPCLOUD, GESDISCCLD like virtual providers. Why is that? 
2. The answer to my question above may already answer this, but the ticket asks that we add all four to the static provider list, why was only LM_FIRMS added?

tle1989:
> Couple general questions:
> 
> 1. What is the difference between providers and virtual providers? I see that we are treating LM_FIRMS like a provider and LARC_CLOUD, LPCLOUD, GESDISCCLD like virtual providers. Why is that?
> 2. The answer to my question above may already answer this, but the ticket asks that we add all four to the static provider list, why was only LM_FIRMS added?

1. The difference is that virtual providers are only used to link to the actually providers.  When you search for arc provider, all the records for that particular provider, and also all the records for those virtual that links to that providers pop up. 
2. Scott checked out with ARC team to figure it out, which are arc providers, which are virtual arc provider; thus, on LB_FIRMS is arc provider, the rest are virtual arc providers.

sritz-23:
> Couple general questions:
> 
>     1. What is the difference between providers and virtual providers? I see that we are treating LM_FIRMS like a provider and LARC_CLOUD, LPCLOUD, GESDISCCLD like virtual providers. Why is that?
> 
>     2. The answer to my question above may already answer this, but the ticket asks that we add all four to the static provider list, why was only LM_FIRMS added?

Good question. This is the explanation from the ARC Team: "think this relates to the DAAC filter in the CMR Dashboard. For example, NSIDCV0 and NSIDC_ECS show up as simply "NSIDC" in the filter"

sritz-23 (approved):
Reviewed changes and approve.

htranho (approved):

PR #462, opened 2024-04-25
CMRARC-813:  Want the Metadata Curation Dashboard to Provide a CSV Report for a Subset of Fields That Have Feedback Provided

mandyparson:
***Need to pull CMRAARC-813 in order to test. Sorry for the typo***

Previously, when a user went to CMR Dash >> In ARC Review Records >> Made a selection and clicked "Get Review Report", they would be given only one option for Exporting Metrics. This button, when clicked, automatically downloads a .csv file with every field for collections and their associated granules in a single line. 

Screenshot 2024-04-25 at 2 01 48 PM

This ticket creates a solution so that a user may be able to "export all metrics" or "export only metrics with feedback". The Export All Metrics button works as the old Export Metrics did while the Export Only Metrics with Feedback button sifts through records and only exports fields that contain feedback (regardless of color flag). 

Screenshot 2024-04-25 at 2 46 42 PM


app/views/reports/review.html.erb
```diff
@@ -5,8 +5,12 @@
     <div class="report_drop_down">
     </div>
     <div class="report_buttons">
-      <%= button_to reports_review_path(format: :csv), params: {record_id: @records.map(&:id)} , data: { turbolinks: false }, class: "report_button" do %>
-        <p>EXPORT METRICS</p>
+      <%= button_to reports_review_path(format: :csv), params: {record_id: @records.map(&:id), full_report: true} , data: { turbolinks: false }, class: "report_button" do %>
+        <p>EXPORT ALL METRICS</p>
+        <i class="fa fa-file-excel-o"></i>
+      <% end %>
+      <%= button_to reports_review_path(format: :csv), params: {record_id: @records.map(&:id), full_report: false} , data: { turbolinks: false }, class: "report_button" do %>
```
cgokey, line 12: I think it is worth figuring out how to do a modal.   It is a pretty big button :)

mandyparson: I confirmed with Tyler and Scott and this solution is adequate

tle1989 (approved):

cgokey (approved):
I still think it is a big button, but given no further comments on it, I'll defer to others :)   We can also add the modal later if we end up adding more buttons.   
Approved.

PR #468, opened 2024-05-06
CMRARC-850: Added shakapacker to rails 7.

htranho:


htranho:
Solved in CMRARC-850b

PR #476, opened 2024-05-09
CMRACR-843: Formatting for umm_json use case

mandyparson:
Addressing an issue found in this ticket where umm_json was not accurate for the format. This is a specific use case where the format of umm_json files is actually umm-c. You can see this issue previously addressed in application_helper.rb line 132:

  def get_format_and_version(format, version)
    version = 'n/a' if version.nil?
    format == 'umm_json' ? "umm-c; version=#{version}" : format
  end

Before, this format said umm_json in the format column, this fix changes it so that it says umm-c. (Chris, I couldn't use the above function because it would have included the version # as well)

Screenshot 2024-05-09 at 1 54 43 PM

### DIFF
--- a/app/controllers/records_controller.rb
+++ b/app/controllers/records_controller.rb
@@ -100,19 +100,13 @@ class RecordsController < ApplicationController
 
     response_array = []
     response_records.each do |record|
-      if record.format == 'umm_json'
-        response_array.push({"id":record.id, "state":record.state, "concept_id": record[:concept_id],
-        "date_ingested": record[:date_ingested], format: 'umm-c',
-        "revision_id": record.revision_id, "short_name": record[:short_name],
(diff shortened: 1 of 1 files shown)


app/controllers/records_controller.rb
```diff
@@ -100,12 +100,21 @@ def find_records_json
 
     response_array = []
     response_records.each do |record|
-      response_array.push({"id":record.id, "state":record.state, "concept_id": record[:concept_id],
-        "date_ingested": record[:date_ingested], "format": record.format,
+      if record.format == 'umm_json'
```
cgokey, line 103: I think of instead of repeating the entire "push", just create a variable called format:
format = record.format == 'umm_json' ? 'umm-c' : record.format
Then just call push with "format" variable instead.

cgokey (approved):

tle1989 (approved):

PR #478, opened 2024-05-13
CMRARC-443: Change Message Viewed by User When Not Provisioned

tle1989:
Currently, when a non-provisioned user tries to login to the Dashboard, they see the following message:

Screenshot 2024-05-13 at 3 32 52 PM

The new change is to update message which is displayed when a non-provisioned user attempts to login to the Dashboard.

Screenshot 2024-05-13 at 3 07 53 PM

### DIFF
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -23,7 +23,7 @@ class LoginController < Devise::OmniauthCallbacksController
     else
       redirect_to root_path
       if (@user.role.nil?)
-        flash.notice = "User is not provisioned with the proper ACLs.   Please contact the Earthdata Operations team."
+        flash.notice = "The user is not provisioned with the proper ACLs. Please contact User Support at support@earthdata.nasa.gov."
       else
         flash.notice = "Could not authenticate from Earthdata Login"
       end
--- a/test/controllers/login_controller_test.rb
+++ b/test/controllers/login_controller_test.rb
@@ -50,13 +50,28 @@ class LoginControllerTest < ActionController::TestCase
       end
 
       it "should set flash notice when user role is nil" do
-        mock_normal_edl_user
 
-        User.any_instance.stubs(:role).returns(nil)
+        AclDao.any_instance.stubs(:get_role_and_daac).returns([nil, nil])
 
         post :urs, params: { provider: :urs }
 
         assert_equal "The user is not provisioned with the proper ACLs. Please contact User Support at support@earthdata.nasa.gov.", flash[:notice]
+      end
+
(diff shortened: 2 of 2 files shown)


app/controllers/login_controller.rb
```diff
@@ -23,7 +23,7 @@ def urs
     else
       redirect_to root_path
       if (@user.role.nil?)
-        flash.notice = "User is not provisioned with the proper ACLs.   Please contact the Earthdata Operations team."
+        flash.notice = "The user is not provisioned with the proper ACLs. Please contact User Support at support@earthdata.nasa.gov."
```
cgokey, line 26: No test breakages?  Do we have a test for this condition?  If not, I think we should write one.   I think the error is occurring when a dashboard is attempting to retrieve ACLs and fails (least I think that is what is happening).

tle1989: No Test breakages; however just added 1 new test based on the feedback.

test/controllers/login_controller_test.rb
```diff
@@ -48,6 +48,16 @@ class LoginControllerTest < ActionController::TestCase
...
         assert(after_count.must_equal(before_count + 1))
       end
+
+      it "should set flash notice when user role is nil" do
+        mock_normal_edl_user
+
+        User.any_instance.stubs(:role).returns(nil)
+
+        post :urs, params: { provider: :urs }
+
```
cgokey, line 58: The issue with the test is it doesn't really test the flow of an actual error coming from CMR.   I think what you want to do is stub the ACL request to CMR with an error, then the code coverage will cover the branch of lines down to this error and you won't need to stub :role.

test/controllers/login_controller_test.rb
```diff
@@ -48,6 +48,31 @@ class LoginControllerTest < ActionController::TestCase
         after_count = User.count
         assert(after_count.must_equal(before_count + 1))
       end
+
+      it "should set flash notice when user role is nil" do
+
+        AclDao.any_instance.stubs(:get_role_and_daac).returns([nil, nil])
```
cgokey, line 54: Again, by stubbing get_role_and_daac, you are not testing the flow of control where CMR returns a response of a user not provisioned.  What happens if you create a EDL user id, try to access dashboard with that user id, and print out the response to the ACL query for your user that is not provisioned.   That response is what you want to mock.

cgokey (approved):

test/controllers/login_controller_test.rb
tle1989: solved

htranho (approved):

PR #481, opened 2024-05-15
CMRARC-844: As an ARC curator, I would like to see some general improvements to the Discussion and Justification section

mandyparson:
Here are all the things this ticket addresses (from ticket AC)

1. Scrollbar is removed from Discussion and Justification section 
2. Making edit and delete icons always visible in the comment section
3. Users have ability to edit both comments and recommendations at the same time and save them _There were a few ideas about how best to do this. However the one we decided on was to change the delete alert to save "any unsaved changes will be lost"_
4. Moving comments up and/or flagging when a comment has been left _I opted to move comments up_
5. Excess scroll bars are removed from comment section 

Before: 
Screenshot 2024-05-16 at 11 31 26 AM

Screenshot 2024-05-16 at 11 32 55 AM

Screenshot 2024-05-16 at 11 31 41 AM

After: 
Screenshot 2024-05-16 at 11 30 49 AM

Screenshot 2024-05-16 at 11 33 39 AM

Screenshot 2024-05-16 at 11 29 59 AM

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1782,6 +1782,8 @@ GEM
     net-smtp (0.5.0)
       net-protocol
     nio4r (2.7.1)
+    nokogiri (1.16.4-arm64-darwin)
+      racc (~> 1.4)
     nokogiri (1.16.4-x86_64-darwin)
       racc (~> 1.4)
     oauth2 (2.0.9)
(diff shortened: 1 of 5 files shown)


tle1989 (commented):
Looks good! except that if the comment on the DISCUSSION & JUSTIFICATION (VISIBLE TO ALL), it covers the New Comment portion since it is at the bottom.  I think it would be better if it is similar to the FEEDBACK DISCUSSION, thus the New Comment should be on top of the old ones.

htranho (approved):

tle1989 (approved):

PR #487, opened 2024-05-21
CMRARC-820: As a user, I can review a umm-c record where the sections appear similar to MMT.

cgokey:
# Overview

### What is the feature?

Reviewing a umm-c record should show sections similar to MMT.

### What is the Solution?

Added the same configuration file that MMT uses for sections and fields to include for each section.   Dashboard now uses this to drive the UI for section/field organization.

Ingest of umm-c records.

# Testing

### Reproduction steps

View any umm-c record in dashboard, you should see the section/field lists have changed, they now match what is in MMT.

### Attachments
Screenshot 2024-05-21 at 5 20 42 PM

# Checklist

- [X] I have added automated tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings

### DIFF
--- a/app/models/record.rb
+++ b/app/models/record.rb
@@ -661,7 +661,7 @@ class Record < ApplicationRecord
     end
   end
 
-  # Returns should return a list where each entry is a (title,[title_list])
+  # Should return a list where each entry is a (title,[title_list])
   def mmt_sections
     section_list = []
 
--- a/app/models/record.rb
+++ b/app/models/record.rb
@@ -626,6 +626,10 @@ class Record < ApplicationRecord
 
   #should return a list where each entry is a (title,[title_list])
   def sections
+    if format == 'umm_json' and recordable_type == "Collection"
+      return mmt_sections
+    end
+
     section_list = []
 
     get_section_titles.each do |title|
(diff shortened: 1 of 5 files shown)


app/models/record.rb
```diff
@@ -657,6 +661,49 @@ def get_section(section_name)
...
+      section = self.get_mmt_section(title, map)
+      if section[0][1].count > 0
+        section_list += section
+      end
+    end
+
+    used_titles = (section_list.map {|section| section[1]}).flatten
+    all_titles = self.sorted_record_datas.map { |data| data.column_name }
+
+    others = [["Other", all_titles.select {|title| !used_titles.include? title }]]
```
tle1989, line 688: What is Other for?

tle1989 (approved):

app/models/record.rb
cgokey: It is other fields not included in the MMT section lists.  It is also a catch all if we ever add fields that are not included in the section lists, they will show up here

PR #493, opened 2024-06-10
CMRARC-865: As an ARC Curator, I would like to see three new dashboard bugs fixed

mandyparson:
# Overview

### What is the feature?

Three issues:
1. The width-resize buttons weren't working
2. update_color_backgrounds() wasn't executing on initial page load
3. Cannot hover over the * to see the acceptable criteria

### What is the Solution?

1. Ended up being a line of .scss called 'content'. Just moved it to another area. Now you can resize the column widths and see the astrick hover-over text. 
2. update_color_backgrounds was too far down the script. It needed to render before other logic came in so I just bumped it up. 

_reviews.scss
the review table when you select a collection >> See Review Detail >> See Collection Review Detail

# Testing

### Reproduction steps

- **Environment for testing: Local
- **Collection to test with: Whichever you'd like

1. Pull down this branch and run rails s
2. Ingest a collection for review if you don't have one
3. Click through to select a collection >> See Review Detail >> See Collection Review Detail
4. You should now be able to re-adjust the column widths and hover over * to see Acceptance Criteria. 
5. If the bubbles were previously colored in, then the color code bar should look like below upon page load (previously it would only block out the cell like that if you clicked on another cell within that row). 

Screenshot 2024-06-10 at 5 11 20 PM
### Attachments

Please include relevant screenshots or files that would be helpful in reviewing and verifying this change.

# Checklist

- [ x] I have added automated tests that prove my fix is effective or that my feature works
- [ x] New and existing unit tests pass locally with my changes
- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] I have made corresponding changes to the documentation
- [ x] My changes generate no new warnings

### DIFF
--- a/app/views/reviews/show.html.erb
+++ b/app/views/reviews/show.html.erb
@@ -189,6 +189,42 @@
 
 
     <script type="text/javascript">
+      //sets all of the flag colors on first/following page loads and updates
+      function update_color_backgrounds() {
+        //hidden select inputs
+        var color_selects = document.getElementsByClassName("color_code_select");
+        //full td cell
+        var color_cells = document.getElementsByClassName("color_code_cell");
+        //the div containing the color selecting icons
+        var color_button_sets = document.getElementsByClassName("color_code_background");
+        //the divs containing the x out option to remove a selected color
+        var color_cancels = document.getElementsByClassName("color_cancel_selection");
+
+        var counter = 0;
+        //iterating over all the hidden color select inputs
+        while (counter < color_selects.length) {
+          var color = color_selects[counter].value
+          //Removing all BG colors from the cell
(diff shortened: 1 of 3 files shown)


tle1989 (approved):

app/views/reviews/show.html.erb
```diff
@@ -189,24 +189,61 @@
 
 
     <script type="text/javascript">
+      //sets all of the flag colors on first/following page loads and updates
```
cgokey, line 192: typically for bugs, if there isn't a test that caught the bug, we should write one.  I don't think you need to do it for the resize of columns (not sure how you'd test that), but this functionality seems testable.

cgokey (approved):

PR #506, opened 2024-07-31
CMRARC-872: Fixed how dashboard calls pyquarc

cgokey:
# Overview

### What is the feature?

For whatever reason, the Faraday library that we were using to call pyquarc no longer works.  I'm not sure if this was due to an upgrade on Impact's team of Pyquarc or maybe a recent security update on our part.  

### What is the Solution?

I switched out library so instead of using Faraday, it now uses HTTParty which is another standard rails library for making REST calls.

Users could no longer ingest records, as the call to pyquarc would fail.

# Testing

Please try ingesting a metadata record into dashboard, prior to this fix, all ingest was failing.

# Checklist

- [X] I have added automated tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings


mandyparson:
After attempting to ingest C1200400820-GHRC in localhost:3000, I get this error (see screenshot)
Screenshot 2024-08-01 at 9 56 47 AM

htranho (approved):

mandyparson:
Now when I click a record to ingest, for example C1200451106-NSIDC_ECS and click "Select Collection" I get this error: 
Screenshot 2024-08-01 at 10 20 47 AM

mandyparson (approved):

